PR #1, opened 2017-01-04
Update to latest version of tech-docs-template

mcmillan:


source/javascripts/start-modules.js
```diff
@@ -1,7 +1,7 @@
 //= require govuk/modules
 //= require modules/anchored-headings
+//= require modules/in-page-navigation
 //= require modules/table-of-contents
-//= require modules/highlighted-nav
```
36degrees: highlighted-nav.js needs deleting.

PR #4, opened 2017-02-03
Postgres encryption and free plan

benh-gds:
Story #138344907 to document free plan and #137716243 to document encrypted plan. Descriptions below

#138344907

What

Document the existence of the Postgresql "free" plan.

Why

Because without authorisation this is the only plan people trying out the PaaS can use, and it is not clear that it exists in the documentation.

If we don't do this

We will continue to receive requests from trial tenants asking for their account to be upgraded when it doesn't need to be,

#137716243

What

Add documentation to the tenant manual about the new Postgres service plans which have "enc" in their name:
database storage is encrypted as rest (when the RDS instance is shutdown)
we recommend that you use it for production workloads
you cannot enable or disable encryption on an existing service
it is not compatible with small plans
we haven't provided it for non-HA plans

Why

We added the plans in #137006863 and users should know about them.

It is done when

There is documentation explaining what it means the encryption, how to use the plans and restrictions.

If we don’t do this

Users might get confused with the new plans, and create DBs without encryption that later cannot be upgraded.

### DIFF
--- a/source/documentation/deploying_services/index.md
+++ b/source/documentation/deploying_services/index.md
@@ -1,19 +1,23 @@
 # Deploy a backing service
 
-## Backing services overview
+## Services and plans
 
-
-Your application probably relies on backing services such as a database, an email delivery service or a monitoring system.
+Many 12-factor applications rely on backing services such as a database, an email delivery service or a monitoring system.
 
 In Cloud Foundry, backing services are referred to as 'services' and are available through the Cloud Foundry ``cf marketplace`` command.
 
 Currently, the only service available is the [PostgreSQL database service](/#postgresql). 
 
-### Paid services
+Each service can have multiple plans available. For example, there are different PostgreSQL plans which vary by availability, storage capacity and encryption.
 
-Some services (including ``postgres``) are considered paid services. Your organization may not have the ability to use paid services enabled. Access to paid services can either be enabled or disabled based on your [organisation quota](/#quotas).
 
-If you try to create a service and receive an error stating "service instance cannot be created because paid service plans are not allowed", please contact us at [gov-uk-paas-support@digital.cabinet-office.gov.uk](mailto:gov-uk-paas-support@digital.cabinet-office.gov.uk).
+### Paid service plans
+
+Some service plans are paid: that is, you can potentially be billed by us for using them. 
+
+By default, access to paid plans is not enabled for a new organisation. Whether this is enabled or not is controlled by your [organisation's quota settings](/#quotas).
+
+If you try to use a paid service and receive an error stating "service instance cannot be created because paid service plans are not allowed", please contact us at [gov-uk-paas-support@digital.cabinet-office.gov.uk](mailto:gov-uk-paas-support@digital.cabinet-office.gov.uk).
 
 ### Accessing services
 
--- a/source/documentation/deploying_services/postgres.md
+++ b/source/documentation/deploying_services/postgres.md
@@ -118,7 +118,7 @@ To create a service and bind it to your app:
 
 Your app must make a [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) connection to the PostgreSQL service. Most libraries use TLS by default.
 
-The Cloud Foundry buildpack will automatically parse the ``VCAP_SERVICES`` [environment variable](/#system-provided-environment-variables) to get details of the PostgreSQL service. The buildpack will set `DATABASE_URL` to the first database found.
+GOV.UK PaaS will automatically parse the ``VCAP_SERVICES`` [environment variable](/#system-provided-environment-variables) to get details of the  service and then set the `DATABASE_URL` variable to the first database found.
 
 Use ``cf env APPNAME`` to see the environment variables.
 
--- a/source/documentation/deploying_services/postgres.md
+++ b/source/documentation/deploying_services/postgres.md
@@ -49,7 +49,7 @@ You should test how your app deals with a failover to make sure you are benefiti
 
 ### Encrypted PostgreSQL plans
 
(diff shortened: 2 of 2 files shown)


source/documentation/deploying_services/index.md
```diff
@@ -1,19 +1,23 @@
...
-
-Your application probably relies on backing services such as a database, an email delivery service or a monitoring system.
+Many 12-factor applications rely on backing services such as a database, an email delivery service or a monitoring system.
 
 In Cloud Foundry, backing services are referred to as 'services' and are available through the Cloud Foundry ``cf marketplace`` command.
 
 Currently, the only service available is the [PostgreSQL database service](/#postgresql). 
 
-### Paid services
+Each service can have multiple plans available. For example, there are different PostgreSQL plans which vary by availability, storage capacity and encryption.
```
mtekel: as of now, all our postgres plans have same, 20GB storage capacity. If we mention it varies, people might ask about it. I'd simply change this to 'capacity' instead of 'storage capacity'. That would then also mean computational/throughput capacity - or any other service specific capacity in general.

source/documentation/deploying_services/postgres.md
```diff
@@ -4,49 +4,62 @@ GOV.UK PaaS enables you to create a PostgreSQL database service (powered by Amaz
...
+During failover, there will be an outage period (from tens of seconds to a few minutes).
+
+See the [Amazon RDS documentation on the failover process](http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.MultiAZ.html#Concepts.MultiAZ.Failover) for more details.
 
-2.  Run:
+You should test how your app deals with a failover to make sure you are benefiting from the high availability plan. We can trigger a failover for you. Please contact us at [gov-uk-paas-support@digital.cabinet-office.gov.uk](mailto:gov-uk-paas-support@digital.cabinet-office.gov.uk) to arrange this.
+
+### Encrypted PostgreSQL plans
+
+Plans with ``enc`` in the name include encryption at rest of the database storage. This means that the data is encrypted while the service is stopped.
```
mtekel: ... 'this means that the data on the disk and in the snapshots is encrypted'.

source/documentation/deploying_services/postgres.md
```diff
@@ -4,49 +4,62 @@ GOV.UK PaaS enables you to create a PostgreSQL database service (powered by Amaz
...
+
+We recommend that you use an encrypted plan for production services.
+
+Once you've created a service instance, you can't enable or disable encryption. There's no way to convert an unencrypted PostgreSQL service instance to an encrypted one later.
```
mtekel: IMO this deserves some highlighting or at least bold text...

source/documentation/deploying_services/postgres.md
```diff
@@ -120,23 +120,25 @@ Your app must make a [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Securit
 
 Your app will need to parse the ``VCAP_SERVICES`` [environment variable](/#system-provided-environment-variables) to get details of the PostgreSQL service (or use a library that does so).
 
-(Note that for some languages/frameworks, the Cloud Foundry buildpack will automatically parse ``VCAP_SERVICES`` and set DATABASE_URL to the first database found.)
+(Note that for some languages/frameworks, the Cloud Foundry buildpack will automatically parse ``VCAP_SERVICES`` and set `DATABASE_URL` to the first database found.)
```
mtekel: This should now be true for _all_ buildpacks.

source/documentation/deploying_services/postgres.md
benh-gds: Following GOV.UK style, we avoid using bold for emphasis. When callout boxes as used on GOV.UK are added to the template, I'll add a warning callout here ( https://github.com/alphagov/govspeak/wiki/Using-govspeak-on-GOV.UK#warning-callouts)

source/documentation/deploying_services/postgres.md
benh-gds: Changed

source/documentation/deploying_services/postgres.md
benh-gds: Removed "Note that for some languages/frameworks,".

benh-gds: On second thoughts, merged with the previous paragraph.

source/documentation/deploying_services/postgres.md
```diff
@@ -118,9 +118,7 @@ To create a service and bind it to your app:
 
 Your app must make a [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) connection to the PostgreSQL service. Most libraries use TLS by default.
 
-Your app will need to parse the ``VCAP_SERVICES`` [environment variable](/#system-provided-environment-variables) to get details of the PostgreSQL service (or use a library that does so).
-
-(Note that for some languages/frameworks, the Cloud Foundry buildpack will automatically parse ``VCAP_SERVICES`` and set `DATABASE_URL` to the first database found.)
+The Cloud Foundry buildpack will automatically parse the ``VCAP_SERVICES`` [environment variable](/#system-provided-environment-variables) to get details of the PostgreSQL service. The buildpack will set `DATABASE_URL` to the first database found.
```
mtekel: technically it's not buildpack doing anything here, but cloud controller. I'd say something like

CF will automatically parse VCAP services to get details of the service and then set DATABASE_URL variable

benh-gds: I'll say "GOV.UK PaaS" instead of CF so it's clearer what we're talking about.

mtekel:
LG, merging.

benh-gds:
Pushed new version to PaaS.

PR #6, opened 2017-02-06
Document SSH environment fix

benh-gds:
Add workaround for setting up SSH environment correctly.

Story #138036037.

To test: check that the documentation now enables SSH sessions to work correctly. Check the command is accurate for all apps. Confirm that it needs to be run every time you SSH (as docs assume currently).

### DIFF
--- a/source/documentation/troubleshooting/ssh.md
+++ b/source/documentation/troubleshooting/ssh.md
@@ -8,15 +8,27 @@ If you do run commands which will change the container temporarily, it's a good 
 
 SSH is enabled by default. In most cases, you will find that you can SSH directly to your app's container.
 
-Simply run:
+1. Run:
 
-```
-cf ssh APPNAME
-```
+    ```
+    cf ssh APPNAME
+    ```
 
-where `APPNAME` is the name of the app. 
+    where `APPNAME` is the name of the app. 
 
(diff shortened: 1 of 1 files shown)


source/documentation/troubleshooting/ssh.md
```diff
@@ -8,15 +8,23 @@ If you do run commands which will change the container temporarily, it's a good
...
+    cf ssh APPNAME
+    ```
+
+    where `APPNAME` is the name of the app. 
+
+2. For some tasks to work, you need to set up the interactive SSH session to match the buildpack environment. To do this, run:
 
-where `APPNAME` is the name of the app. 
+    ```
+    export HOME=/home/vcap/app; [ -d ~/.profile.d ] && for f in ~/.profile.d/*; do source $f; done
```
alext: This doesn't quite match what's specified in the [upstream docs](https://docs.cloudfoundry.org/devguide/deploy-apps/ssh-apps.html#ssh-env). It's missing
```sh
source /home/vcap/app/.profile
```

alext:
This second commit includes several changes that aren't relevant to this PR (changing the URLs etc.) I'm not sure that was intended. It's probably also worth squashing this up into a single commit.

Let me know if you want some help with this.

PR #10, opened 2017-02-15
Document CRLF issue

benh-gds:
Explain to Windows users that pushing with CRLF line breaks can cause
errors.

Story: https://www.pivotaltracker.com/story/show/135421943

## What

Note that Cloud foundry at this time runs all applications in Linux
containers.
Add some instructions to the docs so that new Windows users know that
files with CRLF line endings won't always work

## Suggestions to consider

turn off the autocrlf setting in git
Or make sure that you specify the interpeter to be used in a command in
manifest.yml or a Procfile eg. command: ruby foo.rb

## Why

We had an issue in user research where a user tried to push our sample
ruby app from a windows machine in a git repo with autocrlf turned on.
It failed because Linux was unable to understand the shebang line.

### DIFF
--- a/source/documentation/troubleshooting/line-endings.md
+++ b/source/documentation/troubleshooting/line-endings.md
@@ -1,0 +1,23 @@
+## Push failure from Windows due to CRLF
+
+Pushing an app can fail if you are pushing from a Windows machine that has saved the app files with Windows-style (CRLF) line breaks.
+
+In interpreted languages, you will get a `No such file or directory` error.
+
+This is an example of the error with Ruby:
+
+```
+2016-12-02T06:03:25.95-0800 [APP/PROC/WEB/0]ERR /usr/bin/env: ruby
+2016-12-02T06:03:25.95-0800 [APP/PROC/WEB/0]ERR : No such file or directory
+2016-12-02T06:03:25.95-0800 [APP/PROC/WEB/0]OUT Exit status 127
+```
+
+If you find that pushing an app from a Windows machine is failing like this, you must make sure that files are not being saved with CRLF line breaks. Check the settings in your text editor or IDE.
+
+If the files are from a Git repository, disable Git's `autocrlf` setting:
+
+```
+git config --global core.autocrlf false
+```
+
+then clone the repository again.
--- a/source/documentation/troubleshooting/line-endings.md
+++ b/source/documentation/troubleshooting/line-endings.md
@@ -1,8 +1,8 @@
 ## Pushing apps fails on Windows (line endings issue)
 
-We are aware of some cases where pushing an app can fail if you are pushing from a Windows machine that has saved Cloud Foundry configuration files with Windows-style (CRLF) line endings.
+We are aware of some cases where pushing an app can fail if you are pushing from a Windows machine that has saved project files with Windows-style (CRLF) line endings.
 
-This happens because some buildpacks do not interpret the CRLF line endings correctly. This is an example of an error arising from this problem when using the Ruby buildpack:
+This happens because some buildpacks do not interpret the CRLF line endings correctly. This is an example of an error arising from an executable script with CRLF line endings when using the Ruby buildpack:
 
 ```
 2016-12-02T06:03:25.95-0800 [APP/PROC/WEB/0]ERR /usr/bin/env: ruby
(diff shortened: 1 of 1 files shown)


source/documentation/troubleshooting/line-endings.md
```diff
@@ -0,0 +1,23 @@
+## Push failure from Windows due to CRLF
```
mtekel: do people who are likely to encounter this issue know what 'CRLF' is?

source/documentation/troubleshooting/line-endings.md
```diff
@@ -0,0 +1,23 @@
...
+
+This is an example of the error with Ruby:
+
+```
+2016-12-02T06:03:25.95-0800 [APP/PROC/WEB/0]ERR /usr/bin/env: ruby
+2016-12-02T06:03:25.95-0800 [APP/PROC/WEB/0]ERR : No such file or directory
+2016-12-02T06:03:25.95-0800 [APP/PROC/WEB/0]OUT Exit status 127
+```
+
+If you find that pushing an app from a Windows machine is failing like this, you must make sure that files are not being saved with CRLF line breaks. Check the settings in your text editor or IDE.
```
mtekel: This is not trivial. CRLF are normal line endings on win. It's like asking linux/mac to do CRFL instead of LF...
We could start a flame which ending is correct, some printers also require CRLF for postscript documents.

Moreover, all those interpreted languages, when run on windows directly, work fine with windows line endings...

I wonder if the real issue had been some editor/svn/git adding one more extra line break.

source/documentation/troubleshooting/line-endings.md
```diff
@@ -0,0 +1,23 @@
...
+
+If the files are from a Git repository, disable Git's `autocrlf` setting:
```
mtekel: Not sure I would want to do this if I work on win... I would have to have editor that supports working with linux line endings (and also saving in that format).

mtekel:
In this piece of documentation we say that windows line endings make it unable to push to CF for interpreted languages. For people to work with code on win and deploy to CF, they would need to use editor that uses and saves linux line endings. While this is possible, it's a bit unusual. Same interpreted languages will work correctly locally with win line endings. 

We should be explicit about this (that you need to use editor and in generally work with linux line endings only) or double check that the issue wasn't some tool adding some unusual or extra characters (resulting in e.g. CRLFLF). If this is the case, we should raise issue upstream: CF should support pushing from win. Buildpacks could/should support both types of line endings. It's a matter of simple conversion.

mtekel:
Reading again suggestions to consider: it looks like only CF related files are sensitive to this (manifest, Procfile), the interpreter itself (ruby, python) is able to correctly work with the code. If this is true, we should document that and only require special handling for these files. We should also raise issue with CF upstream, as it should support win properly (as they claim).

source/documentation/troubleshooting/line-endings.md
keymon: This should be clarified, indeed.

source/documentation/troubleshooting/line-endings.md
keymon: Most of the modern programming editors support use Unix line endings, if not all. maybe not notepad, but developers likely use a more advanced editor.

keymon:
CF runs on Linux and linux works with Unix endlines. CRLF can cause random problems, we know by fact that causes issues with the #! of the kernel, but it can cause other problems (e.g. parsing of json, or scripts, config files...)

We should recommend NOT use CRLF, and document the error we do know that  happens (#! in the kernel, that happens with interpreted languages)

benh-gds:
I have asked around on the cross-gov Slack, and it seems that Windows developers on cross-platform projects know that line breaks can cause issues, and will not be outraged by asking to change settings in their editor. I will add a bit more detail explaining that the problem we're aware of is specifically with CF config files, but will keep the general recommendation to avoid CRLF.

keymon:
Sorry for insisting, but if all the discussion was about being clear and explicit, the original affirmation was OK. 

But instead: *"the problem we're aware of is specifically with CF config files"* is not right.

We are aware that executable files of interpreted languages that use the standard `#! ` heading fail. This is how Linux (and most unix AFAIK) work.  CF runs on Linux so it fails executing such files.

keymon:
Good for me.

PR #11, opened 2017-02-15
Reorganise Jenkins doc, document rate limit

benh-gds:
Reorganise Jenkins docs, try to remove impression that using custom
script deployment is less secure. Document issue with login rate limit

https://www.pivotaltracker.com/story/show/136104045

"In our current jenkins docs
(https://github.com/alphagov/paas-tech-docs/blob/01a111ddb7ea95951e2f0d2
c809b317acc292fd5/source/documentation/deploying_apps/jenkins.md), we
recommend users log in every time job runs. This is so that they don't
have to handle secure storage of CF secrets (which is a file with a
logged-in token cf CLI produces after logging in). However our platform
has protection to lock accounts that log in 5 times in 5 minutes. If
someone has a lot of jenkins jobs, or small jobs that run often, they
can easily hit this limit and lock themselves out. As it has already
happened once and support request had to be risen to analyse and
resolve the situation.
We should either document this (that there is a lockout and what
criteria trigger it) or instead think about secure handling of
logged-in credentials and recommend that to users instead."

### DIFF
--- a/source/documentation/deploying_apps/jenkins.md
+++ b/source/documentation/deploying_apps/jenkins.md
@@ -7,13 +7,7 @@ There are two main approaches to pushing applications to GOV.UK PaaS with [Jenki
 
 Both approaches require you to set up the credentials plugin first.
 
-### Login rate limit
-
-For security reasons, the PaaS will lock an account if it logs in 5 times within a 5 minute period. The methods we describe below will trigger a new login every time Jenkins pushes an app. 
-
-If you have multiple Jenkins jobs, or jobs that run frequently, you should take care not to go over this rate limit by having Jenkins push 5 times from the same account within 5 minutes.
-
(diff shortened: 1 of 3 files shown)


mtekel:
I have double checked the login rate limit and there is none, it's only for failed attempts. I think we can remove this paragraph for now from docs. 

On the other side, what is really not nice is that when your account is locked out, you only get 'login failed' message without telling you that you are locked. This then seems like your password is wrong or expired. People then might ask support to reset their PW - whereas in practice, they should have waited 5 mins first and try to log in again. Only then if that fails ask support for new PW.

mtekel:
LG, merging

dcarley:
> On the other side, what is really not nice is that when your account is locked out, you only get 'login failed' message without telling you that you are locked.

I believe it's intentional because it makes it harder to brute force passwords.

PR #14, opened 2017-02-24
[#131250633] More detail about checking service is bound

benh-gds:
#131250633

Provide more details about checking service is bound.
Advise using "restage" rather than "restart" to make sure it binds.

Original story: 

As a user in order to check that an app and Postgres instance have bound successfully, I want to know how to check the service status.
Acceptance Criteria
When a user follows documentation steps to bind app with a backing service, the following step should tell them how to check service status.
User Research
Users felt they only got a ‘proxy answer when checking if app + Postgres instance bound successfully


keymon:
LGTM...

PR #17, opened 2017-02-28
Update to latest version of the template

36degrees:
- Fix a Javascript error on mobile
- Remove extra padding from multi line code blocks
- Remove text from anchor links that’s appearing in Google search results
- Add a .template_version file which should make future updates easier

### DIFF
--- a/.template_version
+++ b/.template_version
@@ -1,0 +1,3 @@
+---
+:remote: https://github.com/alphagov/tech-docs-template.git
+:revision: 22f782e1519eba5d324698059e54b10a1168569e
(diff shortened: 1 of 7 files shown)


dcarley:
@36degrees could you note who should review this and how? Should it be @benh-gds, or anyone on our team?

benh-gds:
I'll test it and merge.

36degrees:
@dcarley I don't think we've really worked the process around this yet. In theory the changes themselves have been reviewed already as a PR against the template. This is really just checking that in updating with the latest version of the template I haven't inadvertently e.g. reverted a change that you'd made.

So the person reviewing should have familiarity with the way the documentation is being _used_. Ben is probably the right person to review at least for now?

dcarley:
That's fine 👍 

Just want to make sure our team is paying attention when they should, and not jumping in when they shouldn't.

PR #19, opened 2017-03-06
Fix errant non breaking space

36degrees:
Replace an errant non breaking space – which causes the heading to be indented – with a normal space.


dcarley:
Verified by running the site locally 👍

PR #20, opened 2017-03-06
[#136353673 #136429263] Remove old URLs from README

dcarley:
## What

Replace all the content about old URLs with the new URL.

The new `docs.cloud` URL was made available in Pivotal story #136353673. The
old `government-paas-developer-docs.readthedocs` URL was redirect to
`docs.cloud` in Pivotal story #136429263.

There is still an outstanding Pivotal story #139284061 to redirect
`paas-tech-docs.cloudapps` to `docs.cloud`, but I think it's sufficient to
remove the notice from the README for now.

## How to review

Check that the new content and my rationale for changing it makes sense.

## Who can review

Probably @benh-gds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,12 +2,7 @@
 
 This is the new technical documentation system for GOV.UK PaaS.
 
-The official docs to give out to tenants are still those hosted at https://government-paas-developer-docs.readthedocs.io/en/latest/ (powered by readthedocs) - however, that will change very shortly.
-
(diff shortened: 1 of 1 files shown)


benh-gds:
I have changed it slightly to mention that the cloudapps.digital URL exists, but should not be used. Just in case someone has the old URL bookmarked or stumbles across it, and is confused by the difference.

dcarley:
Happy with the change 👍 

I think you could squash the two commits before merging though

benh-gds:
Is that something you can do? Whenever I try to squash it just ends up creating more problems.

dcarley:
> Is that something you can do? Whenever I try to squash it just ends up creating more problems.

No problem, done.

Also happy to show you some time if you have any questions.

mtekel:
LG, commits have been squashed, merging.

PR #24, opened 2017-03-14
[#139826733] Restore snapshot experimental, ssh has issues

bleach:
## What

We don't want users to waste time hitting the problems we're currently debugging with SSH tunnels, so I've explained the limitation.

Because we expect people using "restore from snapshot" to use SSH tunnels, and also because we want to get feedback from them, we mark that feature as experimental.

## How to review

Read the docs and form an opinion about if this is clear and accurate for our users.

## Who can review

Anyone but @bleach.

### DIFF
--- a/source/documentation/deploying_services/postgres.md
+++ b/source/documentation/deploying_services/postgres.md
@@ -182,7 +182,7 @@ This feature currently has some limitations:
  * You can only restore the latest snapshot.
  * You cannot restore from a service instance that has been deleted.
  * You must use the same service plan for the copy as for the original service instance.
- * You must create the new service instance in the same organisation and space as the original. This is to prevent unauthorised access to data between spaces. If you need to copy data to a different organisation and/or space, you can use [`pg_dump`](https://www.postgresql.org/docs/9.5/static/backup-dump.html) and [`pg_restore`](https://www.postgresql.org/docs/9.5/static/app-pgrestore.html) via [SSH tunnels](#creating-tcp-tunnels-with-ssh).
+ * You must create the new service instance in the same organisation and space as the original. This is to prevent unauthorised access to data between spaces. If you need to copy data to a different organisation and/or space, you can use [`pg_dump`](https://www.postgresql.org/docs/9.5/static/backup-dump.html) and [`pg_restore`](https://www.postgresql.org/docs/9.5/static/app-pgrestore.html) via [SSH tunnels](#creating-tcp-tunnels-with-ssh). **Note**: it's not currently recommnended to use this for files > 1GB.
 
 To restore from a snapshot:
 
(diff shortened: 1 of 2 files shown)


dcarley:
Makes sense and renders appropriately.

Does it need anyone from content to look at it?

How should we remember when to remove the notice?

bleach:
I think @jollery is going to take a look.

I've added a note to the deskpro ticket linking to this PR and saying we should remove the warning when we fix the problem.

source/documentation/deploying_services/postgres.md
```diff
@@ -170,19 +170,19 @@ Note that data restore will not be available in the event of an RDS outage affec
 
 For more details about how the RDS backup system works, see [Amazon's DB Instance Backups documentation](http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.BackingUpAndRestoringAmazonRDSInstances.html) [external page].
 
-### Restoring a PostgreSQL service snapshot
+### Restoring a PostgreSQL service snapshot (experimental)
 
 As a tenant, you can create a copy of any existing PostgreSQL service instance using the latest snapshot of the RDS instance.
```
jollery: Delete 'As a tenant'

source/documentation/deploying_services/postgres.md
```diff
@@ -170,19 +170,19 @@ Note that data restore will not be available in the event of an RDS outage affec
...
 These snapshots are taken during [the nightly backups described above](#postgresql-service-backup).
 
 This can be useful if you want to clone a production database to be used for testing or batch processing.
 
-This feature currently has some limitations:
+This feature is experimental; we expect it to work and we'd like people who use it to tell us if it worked for them. It has the following limitations:
 
  * You can only restore the latest snapshot.
```
jollery: Change to: 'You can only restore the most recent snapshot from the latest nightly backup'

source/documentation/deploying_services/postgres.md
```diff
@@ -170,19 +170,19 @@ Note that data restore will not be available in the event of an RDS outage affec
...
  * You cannot restore from a service instance that has been deleted.
  * You must use the same service plan for the copy as for the original service instance.
```
jollery: Remove full stops from all bullets from the last, if possible

source/documentation/deploying_services/postgres.md
bleach: This wasn't something that was changed as part of this pull request, but I can do it.

source/documentation/deploying_services/postgres.md
bleach: Again, this wasn't part of this pull request, but it seems small, so I'll do it.

source/documentation/deploying_services/postgres.md
bleach: Again, this wasn't part of my change, but I can do it.

PR #25, opened 2017-03-15
Document docker image deployment support

mtekel:
## What

[Turn on Docker support as experimental](https://www.pivotaltracker.com/n/projects/1275640/stories/140834345)
Add documentation about the feature. Mention limitations.

## How to review

Proofread. Check if formatting and formulations make sense. Check if documentation is complete (e.g. you think it should contain something more) or if instead some parts should be removed.

## Who can review

not @mtekel

### DIFF
--- a/source/documentation/deploying_apps/deploying_docker_image.md
+++ b/source/documentation/deploying_apps/deploying_docker_image.md
@@ -1,0 +1,32 @@
+## Deploy a docker image (experimental)
+
+This section explains how to deploy an app from a docker image. The role of the docker image is to serve as a packaging format and the requirements for the app and its runtime environment are the same as for apps deployed using buildpacks. Configuration via manifest.yml also stays the same, albeit you should not specify buildpack. If you do, it will be ignored, which can create some confusion to anyone reading the manifest. Our platform currently only supports deploying images from docker hub or [Docker Trusted Registry](https://docs.docker.com/datacenter/dtr/2.1/guides/) [external page] servers. You cannot deploy from local images.
+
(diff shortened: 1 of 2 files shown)


alext:
From a quick read-through, this looks good, but we're missing a section explaining the difference in responsibility model from using buildpacks.

mtekel:
Good point, I have added new section.

source/documentation/deploying_apps/deploying_docker_image.md
```diff
@@ -0,0 +1,28 @@
...
+There are few more specifics about docker image support in Cloud Foundry:
+
+* Local storage in each container instance is ephemeral. Every container launch will start with unmodified image.
+* Traffic will be routed to the lowest exposed port (by `EXPOSE` command in Dockerfile). PORT environment variable will be also set to this number. 
+* Buildpack based apps use bash to execute the app start command (`bash -c <command>`), but docker image based apps use sh (`sh -c <command>`) - as bash presence cannot be assumed in your docker image.
+* Privileged container mode is not supported. Features depending on this will not work.
+* Only registries that use Docker Registry API V2 are supported.
+* There is currently no support for deploying images from private/non DTR registries.
+
+### Apps from docker images in production
```
dcarley: We should call this something else, because the current title doesn't reflect the importance and doesn't only apply to production apps. Maybe "Differences in responsibility model" or "Applying security updates"?

mtekel: renamed

combor:
LGTM. Merging.

PR #26, opened 2017-03-15
Stylistic fixes on restore from snapshot docs

bleach:
## What

Made some changes to the language and punctuation suggested by @jollery in #24 - they weren't relevant there and didn't get merged, but didn't want to lose them.

## How to review

- Check that the wording makes sense
- Ensure that the content looks correct in our docs template

## Who can review

Anyone but @bleach

### DIFF
--- a/source/documentation/deploying_services/postgres.md
+++ b/source/documentation/deploying_services/postgres.md
@@ -172,16 +172,16 @@ For more details about how the RDS backup system works, see [Amazon's DB Instanc
 
 ### Restoring a PostgreSQL service snapshot (experimental)
 
-As a tenant, you can create a copy of any existing PostgreSQL service instance using the latest snapshot of the RDS instance.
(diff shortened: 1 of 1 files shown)


source/documentation/deploying_services/postgres.md
```diff
@@ -172,16 +172,16 @@ For more details about how the RDS backup system works, see [Amazon's DB Instanc
 
 ### Restoring a PostgreSQL service snapshot (experimental)
 
-As a tenant, you can create a copy of any existing PostgreSQL service instance using the latest snapshot of the RDS instance.
+You can create a copy of any existing PostgreSQL service instance using the latest snapshot of the RDS instance.
```
keymon: I added "As a tenant" to highlight the difference with the previous section, were we, as operators, can restore an instance to a point in time if the tenant requests it.

jollery: I think the distinction is clear from a content perspective and that 'As a tenant' isn't needed.

PR #31, opened 2017-03-22
Remove docker and route services from limitations section

combor:
## What

Remove docker and route services from limitations section as both features are currently enabled.

## How to review

Sanity check

## Who can review

Not @combor


bleach:
This looks fine. 

I have a lingering worry about whether we're linking to anchor tags that we remove/rename, but I think it's unlikely to happen in this case.

PR #32, opened 2017-04-04
[#134175741] Remove password references from getting started

dcarley:
## What

We are replacing passwords with invites in alphagov/paas-cf#868 so we no
longer need to tell people to reset their password after the first login.

## How to review

Check that the new content makes sense. It makes sense to review this at the same time as alphagov/paas-cf#868.

## Who can review

@jollery or @tomskerous may want to review the content.

### DIFF
--- a/source/documentation/getting_started/quick_setup_guide.md
+++ b/source/documentation/getting_started/quick_setup_guide.md
@@ -42,7 +42,7 @@ GOV.UK PaaS uses a hosting technology called Cloud Foundry. As a tenant (that is
 
     Your `USERNAME` is your email address your account was created with.
 
-    You will then be prompted to enter your password, which should have been sent to you by email (subject "Welcome to GOV.UK PaaS").
+    You will then be prompted to enter your password. You set this when you clicked on the invite link in your welcome email.
 
     If your login fails with a message like this:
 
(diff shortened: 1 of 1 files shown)


source/documentation/getting_started/quick_setup_guide.md
```diff
@@ -42,7 +42,7 @@ GOV.UK PaaS uses a hosting technology called Cloud Foundry. As a tenant (that is
 
     Your `USERNAME` is your email address your account was created with.
 
-    You will then be prompted to enter your password, which should have been sent to you by email (subject "Welcome to GOV.UK PaaS").
+    You will then be prompted to enter your password, which you should have set from the invite link in your welcome email (subject "Welcome to GOV.UK PaaS").
```
tomskerous: Suggested simpler wording. 

You will then be prompted to enter your password. You set this when you clicked on the invite link in your welcome email

dcarley: Updated 👍

henrytk:
There is the minor problem that we haven't actually set up users by invite yet, meaning this documentation applies to nobody reading it. A minor chronological inconvenience shall we say.

PR #33, opened 2017-04-11
[#143024853] Write cloud CI documentation 

combor:
## What

Users want to get going with their low-cost CI solutions, and we need them to do it well, so they don't get compromise themselves.

## How to review

Proofreading please, make sure the docs are sane and info is clear.

You can test rendering with middleman locally:

```
bundle install
bundle exec middleman server
```  

## Who can review
Not @combor


.ruby-version
```diff
@@ -0,0 +1 @@
+2.2.6
```
alext: The `.travis.yml` specifies 2.3.3. This should probably match.

combor: Corrected.

source/documentation/using_ci/index.md
```diff
@@ -0,0 +1,13 @@
+# Configuring a custom continuous integration (CI) system
+
+Setting up a CI service lets you make sure your code changes are delivering new features without introducing new bugs or breaking existing functionality. This helps you to iterate code faster and with greater confidence, a huge benefit once the codebase becomes larger and more complex. 
+ 
+For advanced projects that need high levels of control over configuration, a tool like [Jenkins](https://jenkins.io/) is often used. For simpler projects, or where a team needs to manage deployments of prototypes from their github repositories rather than local code, cloud-based services such as [Travis CI](https://travis-ci.com/) or [CircleCI](https://circleci.com/) are common. Instructions to configure both Jenkins and Travis to run on GOV.UK PaaS are below.
+
+Security is a key concern when setting up a CI system that interacts with a PaaS. You should protect confidential data by: 
+- running your tests in an isolated environment
```
dcarley: I'm not sure I understand this or why it's relevant. Can you explain?

source/documentation/using_ci/index.md
```diff
@@ -0,0 +1,13 @@
...
+- using SSL to encrypt your incoming and outgoing traffic.
```
dcarley: What traffic is this? To and from Cloud Foundry? Won't it always be HTTPS because we enforce it?

source/documentation/using_ci/index.md
```diff
@@ -0,0 +1,13 @@
+# Configuring a custom continuous integration (CI) system
+
+Setting up a CI service lets you make sure your code changes are delivering new features without introducing new bugs or breaking existing functionality. This helps you to iterate code faster and with greater confidence, a huge benefit once the codebase becomes larger and more complex. 
+ 
+For advanced projects that need high levels of control over configuration, a tool like [Jenkins](https://jenkins.io/) is often used. For simpler projects, or where a team needs to manage deployments of prototypes from their github repositories rather than local code, cloud-based services such as [Travis CI](https://travis-ci.com/) or [CircleCI](https://circleci.com/) are common. Instructions to configure both Jenkins and Travis to run on GOV.UK PaaS are below.
```
dcarley: I'm unsure if we should give advice about when to choose Jenkins or Travis. I'm not sure I'd recommend anyone to use Jenkins if they can help it 😁

source/documentation/using_ci/index.md
combor: You are right when it comes to communication between PaaS and CI but that's general advice. For instance traffic between github and travis or between a CI and other internal systems. We can't predict where application source is hosted and what other dependencies it needs to be build and deployed.

source/documentation/using_ci/index.md
combor: We don't want to expose confidential info ( like passwords ) in a shared environment where more people than necessary have access to the build internals.

combor:
> We explicitly do this in the old Jenkins documentation but not in the new overview or Travis documentation. I think we should.

Shall we put it as general advice as it applies to any CI?

source/documentation/using_ci/index.md
tomskerous: I don't think we are recommending it, just saying this is what people do. There are two comon use-cases: Proper TAs wanting to do Proper CI (who might or might not use Jenkins, and will Have Opinions), and designers/tiny teams who just want a quick way to link up deployment to github. As the latter came up in user research, I want something to be able to link out to, in order to reassure those more advanced prototypers.

combor: I'm with Dan on this. It should be up to devs to make a choice base on their needs. We can have guides for most popular ones but I'm unsure about pointing people to a specific one.

tomskerous: I've seen that our trying to be abstract here can actually make people more confused (e.g. they don't know why we refer to 'Cloud Foundry' when they are working with PaaS). And not all our users are *strictly* 'devs'. I'm very happy to be opinionated if it makes things clearer for the most inexperienced users - remember, we have to make it so simple the majority of users succeed first time. But perhaps I'll add a "other CI services are also available" line afterwards.

source/documentation/using_ci/index.md
dcarley: I think it needs to be more specific. Isolating environments for tests is good advice for non-security reasons, but if we feel that it's relevant for *us* to mention in relation to security then we should say why.

Are we recommending something like the following?

- do run your CI tests and then deploy to Cloud Foundry
- do not run your CI tests against Cloud Foundry

combor: TBH I'm not sure.  Do you see any reason we should tell users to do any of these ?

source/documentation/using_ci/index.md
dcarley: I feel like we should be giving advice that's specific to PaaS and this doesn't seem specific enough. Maybe it's also related to my comments on the previous line about assumptions of what they are testing against?

source/documentation/using_ci/index.md
dcarley: What I don't want is for someone inexperienced to choose Jenkins because they think they might need "high levels of control" or their app is "more advanced than a prototype".

I think it's enough to say what they could use - probably in the order of "Travis CI, Circle CI, Jenkins, etc" - and that we have instructions for Travis and Jenkins.

source/documentation/using_ci/index.md
dcarley: Personally I am for removing both of the bullet points and instead having two bullet points about:

- encrypting your secrets (from the paragraph that currently follows the bullet points)
- using a low privilege user

dcarley:
> Shall we put it as general advice as it applies to any CI?

I think so. I wonder if we need to repeat it again in the Travis docs though, to make sure that people don't miss it.

alext:
> Will renaming the pages break any existing links (internal or external)? Is that a problem?

I think it probably will. Having said that, there not much that can be done about this. This is a limitation of the single-page layout this currently uses (we can't do redirects for URL fragments as they're handled in the browser).

dcarley:
The advice about low-privileged users could still be clearer. It doesn't say that it's about your PaaS credentials. It's also easy to overlook if you jump straight to the Travis section. It might be good to have a linkable section in the overview about credentials, which can then be linked to from both the Jenkins and Travis sections.

Discussed with @tomskerous - we're going to merge this because it's required for user research, but will hold the story in review until we update it in a 2nd PR.

PR #34, opened 2017-04-12
[#141117303] Remove references to the API whitelist

Jonty:
## What

We've removed the whitelist, so stop telling people about it.

## How to review

Check that everything referring to the whitelist has been removed, and nothing has been accidentally caught in the crossfire.


Jonty:
@paroxp Fixed!

PR #35, opened 2017-05-01
[#143024853] Improve CI security advice

tomskerous:
Update guidance on use of dedicated accounts for testing, and make
common across Jenkins and Travis.


source/documentation/using_ci/jenkins.md
```diff
@@ -7,6 +7,10 @@ There are two main approaches to pushing applications to GOV.UK PaaS with [Jenki
 
 Both approaches require you to set up the credentials plugin first.
 
+### Set up dedicated users for testing
```
dcarley: I think we should omit "for testing". Most people will deploy from Jenkins "for real" after running tests outside the PaaS.

source/documentation/using_ci/travis.md
```diff
@@ -1,15 +1,19 @@
...
+You will need to know the the particular PaaS `organisation` and `space` you want Travis to deploy your app into. If you don't know, you can get these by logging into PaaS with a terminal window and using the `cf orgs` and `cf spaces` commands.
 
 If you haven't already done so, we also strongly recommend that you add a `name` entry to your `manifest.yml` file so that you'll know the address of your app once published. [Find out about names and URLs on PaaS](#names_routes_domains).
 
+### Plan for dedicated CI accounts
+
+Note that using your own PaaS login details within a CI service should only be a temporary measure. Dedicated user accounts should be set up for testing as soon as reasonably possible - by the end of Alpha at the latest. If you are already an `Org Manager` role on another project in Beta or Live, you should do this anyway. Find out more about [credentials for automated accounts](#credentials-for-automated-accounts).
```
dcarley: Could you describe why this advice is different (and slightly more complex) than the Jenkins advice?

source/documentation/using_ci/jenkins.md
```diff
@@ -7,6 +7,10 @@ There are two main approaches to pushing applications to GOV.UK PaaS with [Jenki
...
+
+In order to avoid developer credentials - particularly those with admin rights - leaking, you should should request a dedicated PaaS user account for use by Jenkins. If you want to deploy to multiple spaces with Jenkins, use a different user account for each. Each user should only have `SpaceDeveloper` access to a single space within your organisation.
```
dcarley: Should this link to `#credentials-for-automated-accounts` (like the Travis section does) rather than duplicating the content?

tomskerous: I thought you were worried that people might miss it.

source/documentation/using_ci/travis.md
tomskerous: Likely to be lower-IA-awareness users, so I'm taking less for granted. They'll want to just get on with it and will use their own accounts at first, but we need them to know it's not sustainable. However, if someone is capable of setting up Jenkins they're much more likely to be happy to wait for dedicated accounts.

There'll be user research on this in due course, so we can carry on improving then.

source/documentation/using_ci/jenkins.md
dcarley: I don't think so? I wrote some thoughts in the previous PR: https://github.com/alphagov/paas-tech-docs/pull/33#issuecomment-297380496

tomskerous: I was thinking about "It's also easy to overlook if you jump straight to the Travis section." But tweaked now.

PR #36, opened 2017-05-03
[#141231851] Add documentation for CDN broker

henrytk:
## What

Add documentation on the new CDN broker service.

## How to review

Check links, readability, and accuracy. It is recommended to use this documentation to test the CDN broker in the corresponding pull request on paas-cf (which has not been raised yet, so this is still technically WIP).

### Before merging

The broker needs to be deployed to production. 

## Who can review

Anyone but @combor, @paroxp, @saliceti or me

### DIFF
--- a/source/documentation/deploying_services/use_a_custom_domain.md
+++ b/source/documentation/deploying_services/use_a_custom_domain.md
@@ -1,0 +1,167 @@
+## Using a custom domain
+
+This section explains how to configure the domain name of your service for production applications.
+
+When you use a custom domain you can also take advantage of the following:
+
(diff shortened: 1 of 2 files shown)


dcarley:
@timmow reminded me that we've been talking about this story as "CDN as a Service" when the epic/feature we're delivering right now is "bring your own domain". Should we tweak the documentation to reflect that? @jollery @tomskerous

dcarley:
Note: @jollery would like to pair on the review of this new documentation. It probably makes sense to do after you have tested #904 though.

henrytk:
@jollery has sent me an updated version of this documentation which I will push up when I get a moment free. Don't merge until then. I will update this comment when it's done.

PR #37, opened 2017-05-04
Change to show only config required to deploy

timmow:
What
====

Change to show only config required to deploy

It confused me that we referenced a complete .travis.yml file, but were
telling the users to add to a .travis.yml file. This changes to list
just the config required to add to .travis.yml file in order to deploy

Who can review
==============
anyone but me


timmow:
This might not be the right solution, but its easier to show what changes I think need to be made as a starting point rather than trying to explain what needs to change

PR #42, opened 2017-06-13
Fix incorrect anchor link

36degrees:
## What

Updated the internal anchor link for the 'Names, routes and domains' section so that it links to a valid hash.

## How to review

Verify that [this is the right place to link to](https://docs.cloud.service.gov.uk/#names-routes-and-domains).

## Who can review

Anyone


dcarley:
Looks good, thanks!

The link checker used by the CI tests has a bug when checking anchors (wummel/linkchecker#557) and I can't see any alternatives that validate anchors 🐼

PR #43, opened 2017-06-13
Add docs on app/service deletion

tomskerous:
Add info on stopping apps, deleting apps and services in the right
order, and also how to delete routes if you forget the -r flag.

Add a line at the end of the “deploy a simple app” in quick setup guide
to remind users they can clean up after themselves.

Content has been reviewed by @bleach while in GDoc form, although there
have been some amendments for clarity once I’ve seen it in context and
formatted.

### DIFF
--- a/source/documentation/deploying_apps/stopping_and_deleting_apps.md
+++ b/source/documentation/deploying_apps/stopping_and_deleting_apps.md
@@ -1,0 +1,52 @@
+## Stopping and deleting apps
+
+It’s important to know that when you perform a ``cf push`` command, there are three things that happen on the platform:
+
+1. your code is uploaded, after which a buildpack converts it to a single container, called a 'droplet', that can be run on PaaS as an app
+2. the droplet is used to start the requested number of instances of that app
+3. a route is created, connecting your app to the internet
+
+When deleting apps you need to remember aspects 1 and 3.
+
+Databases, CDNs and other services have a different lifecycle and need to be removed separately.
+
+### Stopping and starting apps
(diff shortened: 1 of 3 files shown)


source/documentation/deploying_apps/stopping_and_deleting_apps.md
```diff
@@ -0,0 +1,52 @@
...
+which will delete the app and its routes in one go. If your app does have services, please [delete them first](#deleting-services).
+
+If you accidentally delete your app without the ``-r`` option, you can delete the route manually. First confirm the details of the orphaned route by typing the ``cf routes`` command, to get a list of all active routes in the current space. This will list the space, the hostname, the overall domain (normally ``cloudapps.digital``), port, path, type and any bound apps or services. You will see your hostname listed but without an associated app. Use this information to populate the following command:
+
+``cf delete-route [domain name] --hostname [hostname]``
+
+### Deleting services
+Before deleting your app, you will need to delete any services you have provisioned for it.
+
+You can check the details of these by using the cf services command to see all active services in the current space. You will be able to see which services are bound *only* to your app - don’t delete services that other apps are also using.
```
henrytk: Two things here:

1. A minor formatting issue: do we want 'cf services' to be highlighted as code (`cf services`)?
2. It feels like 'You will be able to see which services are bound *only* to your app' is saying when you run `cf services` the result is services which are bound only to your app, which is not true. It returns all services in that space and you can read the output to determine which ones are bound to your app.

PR #45, opened 2017-07-06
Fix broken internal links

klssmith:


dcarley:
Looks good, thank you! 👍 

Unfortunately the automated link checker we use doesn't work with anchors: wummel/linkchecker#557 😞

PR #46, opened 2017-07-07
Merge PostgreSQL and MySQL into Database Services 

jonathanglassman:
## What

Added in MongDB content and updated the Using SERVICE section to apply to all backing services.

## Describe what you have changed and why.

1. Added MongoDB reference to the "Deploy a backing or routing service" > "Services and plans"

2. Grouped High Availability (HA) Plans under another title; "Failover process"
- Specified that HA plans are for Postgres and MySQL
- Added MongoDB content for HA Plans

3. Changed Using PostgreSQL section to Using Database Services, and specified that it applies to PostgreSQL, MySQL and MongoDB. This removes the duplicated content between Using MySQL and Using PostgreSQL and includes MongoDB where appropriate. If this is approved, the separate Using MySQL section would be deleted.

## How to review

- Check if the proposed removal of duplicated content makes sense from a user journey perspective.
- Check if any other information is necessary to enable a user to setup the MongDB backing service.
- Check if the added content on MongoDB HA plans is correct.

## Who can review

Piotr Komborski
Holly Challenger

PR #48, opened 2017-07-11
[#147829023] MySQL: Some libraries don't use TLS by default

dcarley:
## What

We have found that Rails, Django and PHP all don't use TLS by default. It's
not possible to provide a database URI/URL with query parameters that
satisfies most of them. We're going to defer writing more specific
documentation for each individual language until we understand what people
need a bit better. This might result in more support requests in the
meantime.

## How to review

Check if the words make sense.

This can be merged prior to us forcing TLS support.

## Who can review

Anyone.

### DIFF
--- a/source/documentation/deploying_services/mysql.md
+++ b/source/documentation/deploying_services/mysql.md
@@ -132,7 +132,7 @@ To create a service and bind it to your app:
 
 ### Accessing MySQL from your app
 
-Your app must make a [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) connection to the MySQL service. Most libraries use TLS by default.
+Your app must make a [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) connection to the MySQL service. Some libraries use TLS by default, but others will need to be explicitly configured.
 
 GOV.UK PaaS will automatically parse the ``VCAP_SERVICES`` [environment variable](/#system-provided-environment-variables) to get details of the  service and then set the `DATABASE_URL` variable to the first database found.
 


source/documentation/deploying_services/mysql.md
```diff
@@ -132,7 +132,7 @@ To create a service and bind it to your app:
 
 ### Accessing MySQL from your app
 
-Your app must make a [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) connection to the MySQL service. Most libraries use TLS by default.
+Your app must make a [TLS](https://en.wikipedia.org/wiki/Transport_Layer_Security) connection to the MySQL service. Some libraries use TLS by default, but others will need to be explicitly configured.
```
dcarley: For reviewer: do you think it's worth giving any examples here?

dcarley: For reviewer: do you think it's worth saying that the CA certificates are available in `/etc/ssl/certs`?

jonathanglassman:
This can be closed as we have incorporated it into PR #46

dcarley:
> This can be closed as we have incorporated it into PR #46

Discussed offline: #46 is going to be rebased on top of this, because the Mongo docs aren't going to be merged for a few days and we need to get the TLS story finished sooner.

source/documentation/deploying_services/mysql.md
chrisfarms: Probably not inline examples (too much noise), but linking to [example apps](https://github.com/alphagov/paas-cf/tree/master/platform-tests/example-apps)? Which presumably give correctly configured examples.

Possibly this could come later once their are more example combinations to refer to.

PR #49, opened 2017-07-12
MongoDB Beta Documentation

jonathanglassman:
## What

This branch needs rebasing after https://github.com/alphagov/paas-tech-docs/pull/46 is merged. Added in information on self-signed certificates for MongoDB.

## How to review

Read and check it is accurately described.

## Who can review
Dan Carley.

### DIFF
--- a/source/documentation/deploying_services/database_services.md
+++ b/source/documentation/deploying_services/database_services.md
@@ -13,6 +13,7 @@ cf marketplace -s SERVICE
 ```
 
 Where SERVICE is the service you want; it can currently take the following values:
+
  * postgres
  * mysql
  * mongodb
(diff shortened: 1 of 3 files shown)


combor:
This can be reviewed again now.

source/documentation/deploying_services/database_services.md
```diff
@@ -47,37 +51,47 @@ See the [Amazon RDS documentation on the failover process](http://docs.aws.amazo
...
 
-### Encrypted PostgreSQL plans
+#### High availability plans - MongoDB
+
+MongoDB uses replica sets i.e. clusters of MongoDB servers that replicate the contents of the master database. Replica sets provide high availability through sharding, a method for distributing data across multiple machines. A shard is a single mongod instance or replica set that stores some portion of a sharded cluster’s total data set. In the standard configuration for MongoDB on Compose, you get a single-shard with three nodes in a replica-set. Mongo will push all write operations on to the primary data node and then propagate them to the secondary data node on the shard. Should the primary node fail, the secondary will be promoted to primary providing fault-tolerance and redundancy for your databases. The Mongo routers are responsible for dictating any failover, along with load-balancing operations to any of the nodes in the cluster.
```
chrisfarms: > Compose

This is the first and only mention of Compose. Should it be explained that mongo db is backed by Compose's offering? or at the very least this should be a link to Compose to give context.

source/documentation/deploying_services/database_services.md
```diff
@@ -1,41 +1,45 @@
...
 ```
-cf marketplace -s postgres
+cf marketplace -s SERVICE
 ```
 
-Here is a shortened example of the sort of output you will see (the exact plans will vary):
+Where SERVICE is the service you want; it can currently take the following values:
+ * postgres
+ * mysql
+ * mongodb
```
alext: I think `*` is normally fine for lists in Markdown. The problem may be that it needs a blank line before the list.

source/documentation/deploying_services/database_services.md
alext: Do we want to mention Compose at all. So far we've been mostly abstracting that away. We should probably continue to do that.

chrisfarms: I agree.

bleach: In the postgres docs we do mention that it's provided by RDS. This PR removes that, which is not useful to the user IMO. If we're clear about the underlying technology then users can find out more information about them - anything we haven't documented.

But I think this is probably not the correct place for it, the correct place is at the beginning of the Mongo docs

chrisfarms: @jonathanglassman after talking with jess/tom decided to replace "on Compose" with "on our service provider" for now.

We'll go into more detail about the Compose offering in a future docs update.

source/documentation/deploying_services/database_services.md
jonathanglassman: Added in blank line before the list and committed change

source/documentation/deploying_services/database_services.md
jonathanglassman: Replaced "on Compose" with "on our service provider" and committed change

PR #50, opened 2017-07-13
Explain that CI accounts need email addresses

bleach:
## What

It's not currently made obvious that we have a policy that usernames should be
email addresses. One result of this is that users request CI accounts without
including email addresses, the support person has to ask them for an email
address and there's a delay in getting what they need.

This commit makes it clearer that an email address is required when requesting
CI users.

## How to review

- `bundle exec middleman server && open 'http://127.0.0.1:4567/#credentials-for-automated-accounts'`
- proof read text

## Who can review

Anyone but @bleach. Comments welcome from @philandstuff


36degrees:
I think this _might_ be too specific to GDS's use of Google Apps, but is it worth suggesting the use of things like Google Groups for those email addresses?

philandstuff:
👍

bleach:
@36degrees that's a good point. I'll suggest that it might want to be an email address that more than one person can access.

I also missed a point that @philandstuff made in a support ticket that we don't make it clear that CI users are just normal users. I'll try to clear that up too.

bleach:
Pushed a changed commit that addresses those points.

PR #52, opened 2017-07-20
Document how to deploy a Rails 5.1 app with Webpack

tvararu:
## What

I've added a subsection to the "Deploy a Ruby on Rails app" section to document how to deploy a slightly different type of Rails app (that may or may not become prevalent in the near future).

I don't document _why_ `node_modules` have to be excluded: it's because if they're not then `yarn` will not install dependencies but instead use the uploaded ones, which were installed on my Mac, which uses different native extensions compared to Linux x64.

For what it's worth I think this `ln -s .gitignore .cfignore` tip should be much more prominently advertised because it's useful for most applications. 😄  It also really speeds up deploying, so in the quick setup guide I think there could be a step to mention it.

## How to review

Give it a read and see if it makes sense or if it can be simplified / clarified in any way.

## Who can review

Anyone, you shouldn't need Rails specific knowledge to understand what's different about this application type.

### DIFF
--- a/source/documentation/deploying_apps/deploying_rails.md
+++ b/source/documentation/deploying_apps/deploying_rails.md
@@ -135,6 +135,26 @@ You can add another instance of your app by running:
 
 ``cf scale APPNAME -i 2``
 
+### Deploying a Rails 5.1+ app with Webpack
+
+Rails 5.1+ allows you to use [webpacker](https://github.com/rails/webpacker) to install and use node.js modules.
+
+You will need to make the following changes to successfully deploy:
+
+1. Specify multiple `buildpacks` in manifest.yml instead of just one, including both the Ruby and node.js buildpacks from GitHub:
+
+    ```yml
+    buildpacks:
+      - https://github.com/cloudfoundry/ruby-buildpack/
+      - https://github.com/cloudfoundry/nodejs-buildpack/
+    ```
+
+1. Exclude `node_modules` during cloudfoundry deployment, which you can do by following the [#excluding-files](https://docs.cloud.service.gov.uk/#excluding-files) instructions and running:
+
+    ```bash
+    ln -s .gitignore .cfignore
+    ```
+
 ### Web servers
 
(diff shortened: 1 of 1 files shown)


source/documentation/deploying_apps/deploying_rails.md
```diff
@@ -135,6 +135,26 @@ You can add another instance of your app by running:
 
 ``cf scale APPNAME -i 2``
 
+### Deploying a Rails 5.1+ app with Webpack
+
+Rails 5.1+ allows you to use [webpacker](https://github.com/rails/webpacker) to install and use node.js modules.
+
+You will need to make the following changes to successfully deploy:
+
+1. Specify multiple `buildpacks` in manifest.yml instead of just one, including both the Ruby and node.js buildpacks from GitHub:
```
jonathanglassman: Might be worth linking to the buildpacks section of the documentation, just for further information

source/documentation/deploying_apps/deploying_rails.md
```diff
@@ -135,6 +135,26 @@ You can add another instance of your app by running:
...
+      - https://github.com/cloudfoundry/ruby-buildpack/
+      - https://github.com/cloudfoundry/nodejs-buildpack/
+    ```
+
+1. Exclude `node_modules` during cloudfoundry deployment, which you can do by following the [#excluding-files](https://docs.cloud.service.gov.uk/#excluding-files) instructions and running:
+
+    ```bash
+    ln -s .gitignore .cfignore
+    ```
+
```
jonathanglassman: Is it worth putting in some content about how the user will know when they have successfully deployed a rails 5.1+ app with webpack? Or is it unnecessary because it's the same as deploying a rails app, and users would know what success looks like?

camelpunch:
@tvararu it's been a while since there was any activity on this PR. I'm closing this for now, but please re-open if you have time to respond to the feedback @jonathanglassman provided.

PR #53, opened 2017-07-31
[#146274197] Instructions on configuring CDN header forwarding

henrytk:
## What

Instructions on configuring CDN header forwarding.

## How to review

Review in conjuction with the pull request to add this functionality on the paas-cdn-broker repository (hasn't been raised yet).

Proofread. Check the links work. See the repo's README for running instructions.

## Who can review

Anyone but me or @46bit

### DIFF
--- a/source/documentation/deploying_services/use_a_custom_domain.md
+++ b/source/documentation/deploying_services/use_a_custom_domain.md
@@ -81,7 +81,7 @@ Status: create succeeded
 Message: Service instance provisioned [example.com => origin-my-paas-app.cloudapps.digital]; CDN domain d3nrs0916m1mk2.cloudfront.net
 ```
 
-It may take some time for CloudFront to serve your origin from all locations. You can check in advance by hitting CloudFront and passing the custom domain in the `Host` header, using the example below. The CloudFront domain can be found in the output of the command above.
+It may take some time for CloudFront to serve your origin from all locations, because it is a globally distributed network. You can check in advance by sending requests to CloudFront and passing the custom domain in the `Host` header, using the example below. The CloudFront domain can be found by running `cf service my-cdn-route`.
 
 ```
 curl -H "Host: <Custom domain>" https://<CloudFront domain>/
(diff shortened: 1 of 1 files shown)


henrytk:
Hi @NeilBookless 

> May I suggest:
  "The Cloudfront domain can be found in the output from this command:
  [FIRST CODE SAMPLE}

I've pushed up a change to add this. I've put the code example inline though, because the paragraph is already followed by a code sample.

> Q: Why is there a delay? Something like

I've added a part to explain this is because it is a globally distributed network.

> Q: What is the significance of "all locations."? Are some more delayed than others? May be helpful to specify "for example" examples...

This should be addressed by the above point. It's not a latency issue, just the amount of time it takes AWS to provision a globally distributed network.

> Is 'hitting' a widely known technical term or just slang?

It's widely known, but not a technical term so I have changed it to 'sending requests to'

source/documentation/deploying_services/use_a_custom_domain.md
```diff
@@ -105,7 +105,15 @@ After the record is created wait for up to an hour for the CloudFront distributi
 
 #### CDN Configuration Options
```
jollery: I wonder whether this section needs an intro to explain what the contents in the table is/relates to...

source/documentation/deploying_services/use_a_custom_domain.md
```diff
@@ -105,7 +105,15 @@ After the record is created wait for up to an hour for the CloudFront distributi
...
 
+##### Options available
+
+Name | Required | Description | Default
+--- | --- | --- | ---
+`domain` | *Required* | Your custom domain (or domains separated by commas) |
+`cookies` | *Optional* | Forward cookies to the origin | `true` |
+`headers` | *Optional* | A list of headers to forward to the origin | `["Host"]` |
```
jollery: ... and then a section here explaining how tenants should use the table to choose the right option for their needs. We should explicitly state here that more info on each can be found in the relevant sections below.

source/documentation/deploying_services/use_a_custom_domain.md
henrytk: I agree. I've swapped the subheading for a short paragraph explaining what the table is for.

> ... and then a section here explaining how tenants should use the table to choose the right option for their needs. We should explicitly state here that more info on each can be found in the relevant sections below.

I haven't done this because the table doesn't help users do that, but added the reference to the subsections so they know what to read to get the useful information.

henrytk:
@jollery I've addressed your comment, but rebased since so my response may not display in Github. Here's a link to it: https://github.com/alphagov/paas-tech-docs/pull/53#discussion_r131363893

dcarley:
I'll merge this when the deployment has reached production.

PR #55, opened 2017-08-04
Document buildpacks

jonathanglassman:
Added content around standard and custom buildpacks
- New section under Deploying Apps which defines buildpacks and specifies how to use custom buildpacks
- New section, Guidance, at the bottom of the document, which provides around when you should and should not use custom buildpacks, and some advantages of them compared to docker images

This is the first draft and can be iterated; I'll speak to people next week about expanding the content around the responsibilities of supporting a custom buildpack or Docker image, and what responsibilities apply to running an app in production long term.

Please review and see if material and guidance are correct or need any amends or additional content.

Anyone apart from me can review this.

### DIFF
--- a/source/documentation/deploying_apps/buildpacks.md
+++ b/source/documentation/deploying_apps/buildpacks.md
@@ -12,7 +12,7 @@ Cloud Foundry uses buildpacks to provide runtime and framework support for appli
 
 We recommend using the standard buildpacks to maximise the support you will receive from GOV.UK PaaS. Using a custom buildpack or docker image will mean additional setup and maintenance costs.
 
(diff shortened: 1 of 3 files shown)


tomskerous:
Suggest an internal link from "which buildpack should I use" to the guidance section. Otherwise editorially looks good to me. Would be good to get a final tech review for factual accuracy, but getting it published as 90% right for any tenants considering this right now is more important than the final 10%.

jonathanglassman:
Thanks for the quick review; I wanted to put in an internal link right there, but I am not sure how to do so given that the location is not live yet. Will check in Slack.

PR #56, opened 2017-08-07
Add sections for PostgreSQL, MySQL and MongoDB

jonathanglassman:
## What

Have added small sections on each backing service with associated h2 headings. This is to call out the current backing services in the left-hand navigation so that users can scan and quickly see these sections. I have not added them as h3 headings as this would mean editing the config yaml file so that all h3 headings are in the left-hand nav which would not be a good experience for the user. Related to https://www.pivotaltracker.com/story/show/149301683 

## How to review

Check to see if any different or additional content is required in these sections i.e. any specific information that we want to call out on the different backing services.

## Who can review

Anyone but me.


tomskerous:
"powered by Oracle" is a bit marketing-speak otherwise this is good to go.

PR #58, opened 2017-08-09
Deploying an app db services

jonathanglassman:
Update the "Deploying an APP" sections for Ruby, Django and Node.js to reflect that PostgreSQL, MySQL and MongoDB database services.

## How to review

Ensure that the content is accurate.

## Who can review

Anyone apart from Jon Glassman

### DIFF
--- a/source/documentation/deploying_apps/deploying_django.md
+++ b/source/documentation/deploying_apps/deploying_django.md
@@ -2,7 +2,7 @@
 
 This section explains how to deploy an app using the Django framework. You may also need to refer to the [Cloud Foundry documentation about the Python buildpack](https://docs.cloudfoundry.org/buildpacks/python/index.html) [external link].
 
-Note that the only database service currently supported by PaaS is PostgreSQL. If your Django app requires a database, it must be able to work with PostgreSQL.
+Note that the PaaS currently supports the PostgreSQL, MySQL and MongoDB database services. If your Django app requires a database, it must be able to work with one of these.
 
 These steps assume you have already carried out the setup process explained in the [Quick Setup Guide](/#quick-setup-guide) section.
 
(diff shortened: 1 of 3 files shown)


source/documentation/deploying_apps/deploying_django.md
```diff
@@ -2,7 +2,7 @@
 
 This section explains how to deploy an app using the Django framework. You may also need to refer to the [Cloud Foundry documentation about the Python buildpack](https://docs.cloudfoundry.org/buildpacks/python/index.html) [external link].
 
-Note that the only database service currently supported by PaaS is PostgreSQL. If your Django app requires a database, it must be able to work with PostgreSQL.
+Note that the PaaS currently supports the PostgreSQL, MySQL and MongoDB database services. If your Django app requires a database, it must be able to work with one of these.
```
dcarley: Should we link to [#deploy-a-backing-or-routing-service](https://docs.cloud.service.gov.uk/#deploy-a-backing-or-routing-service) as an authoritative list of services rather than duplicating the list in each of these sections? Otherwise it's going to be out-of-date again in the future.

source/documentation/deploying_apps/deploying_rails.md
```diff
@@ -64,9 +64,9 @@ You can add another instance of your app by running:
 
 ``cf scale APPNAME -i 2``
 
-###Deploying with a PostgreSQL database
+###Deploying with a database
```
dcarley: Renaming this section will break link. It doesn't look like we link to it within our own docs though.

The content is still specific about Postgres though. Should we rename it? Does it need some introduction to say that it uses Postgres as an example?

dcarley:
This has been superseded by #74.

PR #59, opened 2017-08-21
[#149301683] Deploying services section re-org

jonathanglassman:
## What

- Re-organise the "Deploy a Backing or Routing Service" section to remove duplicated content.
- Remove duplicated headings within the "Deploy a Backing or Routing Service" section. These duplicated headings were mistakenly implemented when completing previous PR

## How to review

Check that re-organisation has not removed any content inadvertently or caused a change in meaning.

## Who can review

Anyone in PaaS team; have requested Dan and Tom.

### DIFF
--- a/source/documentation/deploying_services/index.md
+++ b/source/documentation/deploying_services/index.md
@@ -1,20 +1,8 @@
 # Deploy a backing or routing service
 
-## Services and plans
-
 Many 12-factor applications rely on backing services such as a database, an email delivery service or a monitoring system. Routing services can be used to proxy and perform preprocessing on application requests such as caching, rate limiting or authentication.
 
-In Cloud Foundry, backing and routing services are referred to as 'services' and are available through the Cloud Foundry ``cf marketplace`` command. GOV.UK PaaS enables you to create a database service and bind it to your app. 
-
-Currently, the following database services are available in the marketplace:
-
- - PostgreSQL
- - MySQL
- - MongoDB (currently only available on private beta)
-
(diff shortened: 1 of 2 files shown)


source/documentation/deploying_services/index.md
```diff
@@ -4,26 +4,37 @@
...
 
+## PostgreSQL
+
+PostgreSQL is an object-relational database management system. It is open-source and designed to be extensible; currently the postgis and uuid-ossp extensions are enabled.
+
+## MySQL
+
+MySQL is an open source relational database management system that uses Structured Query Language (SQL) and is backed by Oracle.
+
+## MongoDB
```
dcarley: This position of these H2 headings makes `### Paid service plans` a child of `## MongoDB` instead of `## Services and plans` which it was previously, which I think is incorrect.

I don't have a good suggestion of how to solve that, because making them H3s would mean that they don't appear in the left-hand navigation, which @tomskerous was really keen on doing.

Any thoughts?

jonathanglassman: I agree on the parent-child headings issue and recognised this at time of drafting, but I also couldn't think at the time of a better way to organise this information given the LH Nav requirement. If we include H3s in the LH Nav, then it will be really long. @tomskerous would you be willing to have the database names as H3s?

dcarley: Tom is off until the 5th. @jollery has agreed to have a think about this tomorrow.

jonathanglassman:
@dcarley made changes required

dcarley:
Looks much better @jonathanglassman  👍 

Thanks for the idea @jollery 👍

PR #64, opened 2017-09-20
[Support] Fix hostname in route-services docs.

alext:
## What

The non-production base domain name had been used here in error.

## How to review

A simple code review is probably enough.

## Who can review

Not me.


46bit:
LGTM. Especially when the line below references `cloudapps.digital`.

PR #66, opened 2017-10-02
[#150879493] Add info about changing plans with update-service

chrisfarms:
## What

This adds a brief note about changing service plans and how long they
might be expected to wait for the changes to take effect. We also
document the apply_at_maintenance_window param which is available for
RDS-broker instances.

## How to review

⚠️  https://github.com/alphagov/paas-cf/pull/1049 should be merged before this.

Proof read, check formatting.

## Who can review

### DIFF
--- a/source/documentation/deploying_services/database_services.md
+++ b/source/documentation/deploying_services/database_services.md
@@ -133,6 +133,29 @@ To create a service and bind it to your app:
     and check the ``Bound apps:`` line of the output.
 
 
+### Changing service plans
+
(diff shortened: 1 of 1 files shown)


alext:
Line 181 says:
> We will add the ability to set the time of the maintenance window in a future version of GOV.UK PaaS.

Does this need to be updated now that changing this is possible?

paroxp:
I've created another commit covering your comment. Ready to be fixed up.

PR #67, opened 2017-10-06
Third Party Log Management

jonathanglassman:
## What

Added content for third party log management services, as per story https://www.pivotaltracker.com/story/show/150847208

## How to review

1. Check that the content is accurate and the structure makes sense
1. Manually check the links work and are appropriate to the content

## Who can review

Any dev in PaaS team.

### DIFF
--- a/source/documentation/troubleshooting/index.md
+++ b/source/documentation/troubleshooting/index.md
@@ -31,7 +31,7 @@ Running ``cf events`` shows you when an app starts, stops, restarts, or crashes 
 
 ### Third-Party Log Management Services
 
-The default Cloud Foundry logging system stores a limited amount of logging information for a defined time. You can retain more logging information for a longer period of time by setting up a third-party log management service. Instructions on how to set up a log management service can be found [here](https://docs.cloudfoundry.org/devguide/services/log-management.html) [external link]. The next section details how to set up an example log management service, [Logit.io](https://logit.io/) [external link].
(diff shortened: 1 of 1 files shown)


source/documentation/troubleshooting/index.md
```diff
@@ -29,12 +29,131 @@ cf events APPNAME
 
 Running ``cf events`` shows you when an app starts, stops, restarts, or crashes (including error codes). The output is often easier to interpret than the output of ``cf logs``.
 
+### Third-Party Log Management Services
+
+The default Cloud Foundry logging system stores a limited amount of logging information for a defined time. You can retain more logging information for a longer period of time by setting up a third-party log management service. Instructions on how to set up a log management service can be found [here](https://docs.cloudfoundry.org/devguide/services/log-management.html) [external link]. The next section details how to set up an example log management service, [Logit.io](https://logit.io/) [external link].
```
46bit: @jonathanglassman I'm wondering if this could be misleading:

> You can retain more logging information for a longer period of time by setting up a third-party log management service.

The 'more' could be confusing: it implies that a third-party system gets more logs than `cf logs --recent` option. But the key benefit is that logs are retained for longer. I wonder if removing the 'more' would work:

> You can retain logging information for a longer period of time by setting up a third-party log management service.

What do you think?

jonathanglassman: I think you're right - I'll remove. Thanks for the review.

46bit:
👍

PR #68, opened 2017-10-10
logit logstash config: clearer parse failure tag

philandstuff:
`_grokparsefailure` is a generic tag that any grok filter will throw
by default.  By setting a tag specific to the particular filter, it
makes it much clearer exactly where in the filter processing the parse
failure happened.

This filter configuration already did this for the router grok
parsing (with `_routerparsefailure`); this commit does the same for
the syslog grok parsing.


philandstuff:
I've manually tested this on paas with logit.  I did the following:

 - checked that app logs are still correctly parsed and tagged
 - checked that if I shove bad data in to logstash via the tcp input, I get a `_syslogparsefailure` tag

I think this is good enough to go.

dcarley:
Thanks. I haven't tested it myself, but it makes sense to me 👍

PR #71, opened 2017-10-17
Bumped the logging services up the hierarchy

tomskerous:
We didn’t have anywhere in the nav or h2s that talked about logging
services. Bumped the H3 to a H2 and added the word “Using” to help it
make more sense in the context.


samcrang:
Presumably everything under this heading also needs to be bumped up one level in the hierarchy too?

paroxp:
I've addressed @samcrang's comments to move this along.

samcrang:
Seems reasonable 👍

PR #72, opened 2017-10-20
[#139841751] Take out non-working http notification

paroxp:
## What

Now that we've implemented a redirect from HTTP to HTTPS for all of our
routes, the lines informing users about apps not working under `http://`
are no longer relevant.

We aim to remove any information which may be false to prevent confusion
of our users.

## How to review

- Check if the change  only removes appropriate lines
- Check if there are no more false information around this topic


bleach:
Should we mention what the new behaviour is? Can people rely on HTTP URLs working?

PR #73, opened 2017-10-23
[#139841751] Add redirection notice and justification

paroxp:
## What

As we have recently implemented a functionality that will redirect any
http traffic to https. It will also strip the URL path along with the
query parameters. We'd like to have few paragraphs in our docs to
explain that we're doing this and why.

## How to review

- Check if the new content makes sense

### DIFF
--- a/source/documentation/deploying_apps/index.md
+++ b/source/documentation/deploying_apps/index.md
@@ -32,6 +32,18 @@ After deployment, you can increase the running instances to two using:
 
 ### Caveats
 * Your app should not write to local storage. Cloud Foundry local storage is ephemeral and can be deleted at any time.
-* You may need to set environment variables for your app to work. All configuration information should be stored in environment variables, not in the code. 
+* You may need to set environment variables for your app to work. All configuration information should be stored in environment variables, not in the code.
 * Instances will be restarted if they [exceed memory limits](/#quotas).
 * Your application should write all its log messages to `STDOUT`/`STDERR`, rather than a log file.
+
+## Secure and non-secure requests
+
+Requests could be made to the non-secure `http://` protocol due to:
+
+ * misconfiguration of an app that allows non-encrypted traffic through
+ * a service linking to the HTTP version of a page by mistake
+
+In this situation, any requests made to the `http://` protocol will be
+automatically redirected to the base HTTPS version of that URL. The original
+query path and query parameters will be removed. This prevents a site from
+repeatedly redirecting back to the HTTP protocol without the user noticing.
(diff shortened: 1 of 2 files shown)


bandesz:
I don't want to be the final reviewer, but wanted to add my notes:
 - I think we should mention in the first paragraph that we are stripping the path and the query parameters - people tend to read selectively and some will not read the second block.
 - for the part why we are doing this: sensitive values in GET/POST parameters are readable in http traffic and stripping the path/query parameters will help developers realise they are using the wrong protocol.

dcarley:
I don't think this should be under the "deploy a static site" section because it applies to all apps. I wonder if it needs a section of its own.

@jonathanglassman would you be able to help with where to put it and rephrasing to make sure it's user friendly?

paroxp:
I know it applies to all apps on the system and I was wondering why did the previous note exist in the static site section only... 

I was thinking of creating a separate section for it, but I'm not sure if it'd be noticeable enough... For instance, if a user comes in, they'd navigate to `Deploying Rails` section and probably leave there. Whilst if it's a `Note` section, they may get a feeling it's more related to themselves. 

I don't really want to duplicate it for every section that deploys an app...

jonathanglassman:
I agree with Dan and Rafal, I think a separate section under the Deploying apps section would be a good way to structure this content to make it more noticeable. That way it will appear in the LH nav and be scannable, and would hopefully mean that a user would notice it even if they go straight to the relevant app section. @paroxp, maybe we could have a quick chat on this? I just want to make sure I understand it correctly.

paroxp:
We had paired on this one with @jonathanglassman and made it more approachable.

We decided to move it into `Deploying Apps` > `Secure and non-secure requests` which makes it generic enough and creates a side menu link. We weren't sure of the title, so please guide if you dislike it.

Ready for re-review.

source/documentation/deploying_apps/index.md
```diff
@@ -32,6 +32,19 @@ After deployment, you can increase the running instances to two using:
...
+## Secure and non-secure requests
+
+Requests could be made to the non-secure `http://` protocol due to:
+
+ * misconfiguration of an app that allows non-encrypted traffic through
+ * a service linking to the HTTP version of a page by mistake
+
+In this situation, any requests made to the `http://` protocol will be
+automatically redirected to the base HTTPS version of that URL. Any information
+that is sent as part of that request will be blocked. This ensures that the
```
dcarley: I'm unsure about this because it could feel a bit ambiguous to people that understand HTTPS well:

> Any information that is sent as part of that request will be blocked. 

What do you think about the following?

    The original query path and query parameters will be removed.

Similarly for:

> This ensures that the connection between the client and the server is secure and all information is encrypted.

The original request will have still happened over HTTP. I'm not sure how to represent what @bandesz said about us redirecting to make people aware.

paroxp: Addressed after conversation with Jon and Dan on Hangouts.

dcarley:
Thanks again @paroxp @jonathanglassman 👍

PR #74, opened 2017-10-26
[#152315819] Backing service update

jonathanglassman:
## What

Updated tech docs to reflect that multiple backing services are supported by PaaS rather than just PostgreSQL.

## How to review

1. Check that content makes sense and is consistent with https://docs.google.com/document/d/1Ilpx4Ekibo0o3mT0xqWyFus2tqxrBE7scmj_bxC2yBI/edit#

## Who can review

Anyone except Jon Glassman

### DIFF
--- a/source/documentation/deploying_apps/deploying_django.md
+++ b/source/documentation/deploying_apps/deploying_django.md
@@ -2,7 +2,7 @@
 
 This section explains how to deploy an app using the Django framework. You may also need to refer to the [Cloud Foundry documentation about the Python buildpack](https://docs.cloudfoundry.org/buildpacks/python/index.html) [external link].
 
-Note that the only database service currently supported by PaaS is PostgreSQL. If your Django app requires a database, it must be able to work with PostgreSQL.
+> If your app requires a [backing service](https://docs.cloud.service.gov.uk/#deploy-a-backing-or-routing-service), it must be able to work with one of the services supported by PaaS. Instructions for both are given in this section.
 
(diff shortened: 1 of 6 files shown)


source/documentation/deploying_apps/deploying_django.md
```diff
@@ -2,7 +2,7 @@
 
 This section explains how to deploy an app using the Django framework. You may also need to refer to the [Cloud Foundry documentation about the Python buildpack](https://docs.cloudfoundry.org/buildpacks/python/index.html) [external link].
 
-Note that the only database service currently supported by PaaS is PostgreSQL. If your Django app requires a database, it must be able to work with PostgreSQL.
+> If your app requires a [backing service](https://docs.cloud.service.gov.uk/#deploy-a-backing-or-routing-service), it must be able to work with one of the services supported by PaaS. Instructions for deploying both backing service and non-backing service apps are given in this section.
```
dcarley: Could you change all of the links in this PR to be root-relative rather than absolute? For example `/#deploy-a-backing-or-routing-service` instead of `https://docs.cloud.service.gov.uk/#deploy-a-backing-or-routing-service`. This will ensure that the links work when the site is deployed somewhere else, such as previewing locally.

jonathanglassman:
@dcarley I've made the changes requested; hope that's ok?

dcarley:
Thanks for updating 👍 

Some of these commits could be squashed together, such as the "Update as per TD feedback" commits, because they don't provide much additional information to someone looking at the commit history. But I don't think it's worth holding up the review of this particular story any further so I'm going to merge as-is.

PR #76, opened 2017-10-31
update to reflect cf buildpacks command

gidsg:
## What
Update documentation to reflect that `cf buildpacks` can be used to check installed buildpacks and version numbers.

## How to review


dcarley:
This is good, thanks 👍

PR #81, opened 2017-11-15
Check ~/.profile exists before sourcing it

sakisv:
## What

Updated one of the commands a user needs to run to setup their environment when ssh-ing onto an instance to be more defensive.

The reason for the change is that bash complains that the file doesn't exist, which, although not a big issue, it could easily be avoided.

## How to review

1. Copy and paste the new command after you ssh'd onto an instance to ensure that existing functionality didn't break.

## Who can review

Anyone.


alext:
Thanks for this.

PR #82, opened 2017-11-16
[#152231536] Upgrade to the latest tech-docs template

alext:
## What

Updated with:
```
bundle exec middleman init . -T alphagov/tech-docs-template
```

~~~One effect of these changes is that the `hr`s between sections are now gone. This is a change in the upstream template. If this proves to hurt usability, we can look at manually adding them back in a follow-up PR.~~~

Update: I've now added a commit to restore the separator lines. I intend to raise a PR against the template to restore them there.

## How to review

Preview the content locally ([see README](https://github.com/alphagov/paas-tech-docs#preview)) and check that it still renders as expected.

## Who can review

Not me.


46bit:
LGTM. Most of the changes to the template seem like good improvements. Line spacing for body text has been reduced, which makes our larger paragraphs look a bit more like walls of text, but this doesn't appear a major issue.

.template_version
```diff
@@ -1,3 +1,3 @@
 ---
 :remote: https://github.com/alphagov/tech-docs-template.git
-:revision: 22f782e1519eba5d324698059e54b10a1168569e
+:revision: head
```
46bit: Do we know if it matters that this doesn't point to a specific commit?

alext: I'm not sure it does. I followed the instructions in the tech-docs-template repo for updating this, and this is the changes it made.

This isn't using submodules or anything like this, so I don't know what this is actually used for. When I updated the template, it pulled down all the files from the upstream repo, and wrote them into this repo prompting for any conflicts etc.

46bit: I tried to find the relevant code in Middleman but Github search came up empty. Running the upgrade command on top of `head` worked fine so let's not worry about this.

PR #83, opened 2017-11-16
[#152638275] Add separate GOV.UK crown logo image for PaaS HTML emails

bandesz:
## What

I copied source/images/gov.uk_logotype_crown.png as a separate file to be
used in the PaaS HTML email headers. This way we can avoid breaking the emails
if the original logo is changed.

## How to review

After merge please check if https://www.cloud.service.gov.uk/images/email/gov.uk_logotype_crown.png becomes available.

## Who can review

Not @bandesz


dcarley:
This repo serves https://docs.cloud.service.gov.uk/ whereas it's [alphagov/paas-product-page](https://github.com/alphagov/paas-product-page) that serves https://www.cloud.service.gov.uk/

I think it's OK to use the existing URL in the product pages, but we should add a test that asserts it's there and describes the additional usage so that people get feedback if we change it.

alext:
> After merge please check if https://www.cloud.service.gov.uk/images/email/gov.uk_logotype_crown.png becomes available.

Did you mean docs.cloud.service.gov.uk? That's where this site is deployed. If you want www.cloud.service.gov.uk you'll need to update alphagov/paas-product-page.

bandesz:
After talking with @alext I'll close this PR as we're going to use the existing logo from the paas-product-page. I'll add a test to that repo to make sure we don't delete that specific image.

PR #84, opened 2017-11-16
Update Logstash filter to create fields for app/space/org

davbo:
## What
Update Logstash filter to create fields for app/space/org

We found syslog_host contained these fields but it caused some confusion as it
isn't the most obvious place to find the app name.

Creating separate fields for these will hopefully be more intuitive and make it
easier to produce queries when dealing with multiple apps shipping logs to a
single ELK stack.

## How to review

1. Apply this configuration to an ELK stack
2. Ship logs to that ELK stack from CloudFoundry
3. Look in Kibana for the app / space / org fields

## Who can review

Anyone familiar with logstash configuration.

### DIFF
--- a/source/documentation/troubleshooting/index.md
+++ b/source/documentation/troubleshooting/index.md
@@ -72,6 +72,13 @@ filter {
         }
     }
 
+    # Cloud Foundry passes the app name, space and organisation in the syslog_host
+    # Filtering them into separate fields makes it easier to query multiple apps in a single Kibana instance
+    dissect {
+        mapping => { "syslog_host" => "%{[cf][org]}.%{[cf][space]}.%{[cf][app]}" }
+        tag_on_failure => ["_sysloghostdissectfailure"]
+    }
+
     # Cloud Foundry gorouter logs
     if [syslog_proc] =~ "RTR" {
         mutate { replace => { "type" => "gorouter" } }


davbo:
Thanks to @philandstuff for helping me group the fields into a `cf` document.

gidsg:
👍 Have tested on Registers and works as expected.

dcarley:
Thanks! Someone from our team will review this soon.

source/documentation/troubleshooting/index.md
```diff
@@ -72,6 +72,12 @@ filter {
         }
     }
 
+    # Cloud Foundry passes the app name, space and organisation in the syslog_host
+    # Filtering them into separate fields makes it easier to query multiple apps in a single Kibana instance
+    dissect {
+        mapping => { "syslog_host" => "%{[cf][org]}.%{[cf][space]}.%{[cf][app]}" }
```
philandstuff: hmm, is it worth specifying a [`tag_on_failure`](https://www.elastic.co/guide/en/logstash/current/plugins-filters-dissect.html#plugins-filters-dissect-tag_on_failure)?  By default this will tag `_dissectfailure`, but I've found it useful in the past to have a unique tag per filter to make it clear where in the filter configuration the failure happened.

Of course, right now this is the only `dissect` filter in the config, but that may well change.

I propose something like `tag_on_failure => ["_sysloghostdissectfailure"]` although naming is hard.

samcrang: @philandstuff @davbo I'm going to make Phil's proposed change and then get this merged

PR #86, opened 2017-12-04
Metrics section title change

jonathanglassman:
## What

"Exporting metrics data" changed to "Metrics" as per feedback from Michael Mokyrz. This makes it consistent with Logs.

## How to review

1. Review title to make sure it makes sense.

## Who can review

Recommend @46bit or @samcrang.


bleach:
This broke the link from https://github.com/alphagov/paas-cf-apps-statsd/blob/f7f21bacce6244eeafb11d108e6eec1f320b392c/README.md#getting-started

PR #89, opened 2017-12-08
[#148415719] Add docs about tracing requests to troubleshooting (Zipkin)

bandesz:
## What

Zipkin was enabled in the Cloud Foundry router which allows developers to trace HTTP requests and responses. We want our tenants to know about this feature.

## How to review

Read the Troubleshooting/Tracing requests paragraph and check if it makes sense.

## Who can review

Not @bandesz

### DIFF
--- a/source/documentation/troubleshooting/tracing.md
+++ b/source/documentation/troubleshooting/tracing.md
@@ -1,0 +1,9 @@
+## Tracing HTTP requests
+
+The Cloud Foundry router automatically adds Zipkin-compatible headers to the incoming HTTP requests so you can match those with your application's responses.
+
+> Zipkin is a tracing system that enables app developers to troubleshoot failures or latency issues. Zipkin provides the ability to trace requests and responses across distributed systems. See [Zipkin.io](https://zipkin.io) [external link] for more information.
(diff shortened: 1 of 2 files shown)


source/documentation/troubleshooting/tracing.md
```diff
@@ -0,0 +1,9 @@
+## Tracing HTTP requests
+
+The Cloud Foundry router automatically adds Zipkin-compatible headers to the incoming HTTP requests so you can match those with your application's responses.
+
+> Zipkin is a tracing system that enables app developers to troubleshoot failures or latency issues. Zipkin provides the ability to trace requests and responses across distributed systems. See [Zipkin.io](https://zipkin.io) [external link] for more information.
+
+To trace app requests and responses in Cloud Foundry, apps must also log Zipkin headers.
```
henrytk: Do we need to elaborate on how they would do this?

bandesz: I don't think so, because it's about literally logging the Zipkin headers. This whole feature should be mostly for people who know what Zipkin is or they willing to learn it.

PR #93, opened 2017-12-14
[#152632922] Remove file extensions from product page links

dcarley:
## What

Links to the product pages shouldn't have a file extension because
Middleman renders them without and that's how they are referenced in the
product pages themselves.

Most of these are in the header, but there were also some references to the
support page within the custom domain documentation.

These also caused the requests to redirect off of `service.gov.uk` and go to
`cloudapps.digital` because of a bug that we're fixing in a separate story.

## How to review

1. Follow all of the links in the top-right header and check that they work.
1. Follow all of the support links in [/#using-a-custom-domain](http://localhost:4567/#using-a-custom-domain) and check that they work.

## Who can review

Anyone.

### DIFF
--- a/config/tech-docs.yml
+++ b/config/tech-docs.yml
@@ -10,10 +10,10 @@ phase: Beta
 # Links to show on right-hand-side of header
 header_links:
   About: https://www.cloud.service.gov.uk
-  Features: https://www.cloud.service.gov.uk/features.html
-  Roadmap: https://www.cloud.service.gov.uk/roadmap.html
+  Features: https://www.cloud.service.gov.uk/features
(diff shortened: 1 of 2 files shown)


alext:
Changes look good, I'm not sure the commit message is correct though. The product pages are no longer rendered by middleman - they're now a Sinatra app. I believe it was this change that changed the URLs.

dcarley:
> Changes look good, I'm not sure the commit message is correct though. The product pages are no longer rendered by middleman - they're now a Sinatra app. I believe it was this change that changed the URLs.

ACK, removed the referenced and rebased.

PR #100, opened 2017-12-21
[#153744593] More than one instance

jonathanglassman:
## What

Changed "at least two app instances" to "more than one app instance" so that tenants do not think they just have to run two instances. Also added detail on why they should do this.

## How to review

1. Check the content addresses the story https://www.pivotaltracker.com/story/show/153744593

## Who can review

Anyone except Jon Glassman. Dan Carley preferred as he wrote the original story.

### DIFF
--- a/source/documentation/managing_apps/scaling.md
+++ b/source/documentation/managing_apps/scaling.md
@@ -12,10 +12,12 @@ from the command line.
 
 If you are anticipating a spike in demand for a service hosted on GOV.UK PaaS, please contact us well in advance at [gov-uk-paas-support@digital.cabinet-office.gov.uk](mailto:gov-uk-paas-support@digital.cabinet-office.gov.uk).
 
-###Increasing instances
+### Increasing instances
 
-You can change the number of instances of your app running in parallel.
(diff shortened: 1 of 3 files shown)


source/documentation/managing_apps/scaling.md
```diff
@@ -12,10 +12,12 @@ from the command line.
...
 
-###Increasing instances
+### Increasing instances
 
-You can change the number of instances of your app running in parallel.
-Incoming requests are automatically load-balanced across all instances.
+You can change the number of instances of your app running in parallel. Running more than one app instance:
+
+- allows your app to handle increased traffic and demand as incoming requests are automatically load-balanced across all instances
+- helps maintain high availability and decreases the likelihood that the failure of a cell will take down your app
```
chrisfarms: > failure of a cell

Do we mention the concept of "cells" anywhere else? ... this might be a confusing new term.

Maybe:

> helps maintain high availability by decreasing the likelihood that the failure of a single component could take down your app

jonathanglassman: agree, we don't mention cell anywhere else in the docs so component is an acceptable generic term

PR #108, opened 2018-01-18
[#143584125] Add link to self-service password reset

chrisfarms:
# What

We have enabled the self service password reset feature and want to
bring it to the attention of tenants who may be looking for what to do if they have fotgotten their password.

# How to review

* Check it makes sense
* Preview it `bundle exec middleman server` 

## Who can review

not @chrisfarms or @henrytk

### DIFF
--- a/source/documentation/troubleshooting/password.md
+++ b/source/documentation/troubleshooting/password.md
@@ -1,0 +1,5 @@
+## Forgotten passwords
+
+If you believe you have forgotten your GOV.UK PaaS credentials, you can request a password reset using the link below:
+
+[Reset your paswword](https://login.cloud.service.gov.uk/forgot_password)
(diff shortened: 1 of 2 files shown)


source/documentation/troubleshooting/password.md
```diff
@@ -0,0 +1,5 @@
+## Forgetten passwords
```
keymon: `s/Forgetten/Forgotten/`

henrytk: Fixed. The link was also wrong. It now points to `/forgot_password`, not `/password_reset`.

keymon:
:+1: LGTM 

Shall we wait for the feature to be in live before merging the docs?

henrytk:
> Shall we wait for the feature to be in live before merging the docs?

Yes

PR #112, opened 2018-02-10
Add a missing word in privacy.md

issyl0:
- There were multiple options for what to do, either removing "to" or adding "store". I went with the latter as it made more sense with "on servers outside". I'm happy to be corrected though!
- My text editor also added a trailing new line to the end of the file.

### DIFF
--- a/source/documentation/getting_started/privacy.md
+++ b/source/documentation/getting_started/privacy.md
@@ -33,7 +33,7 @@ We will delete your personal data when your user account is no longer required. 
 
 ### Where we store your data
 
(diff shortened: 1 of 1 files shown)


samcrang:
@issyl0 thanks for this! I'll try and find out who will know what this was meant to say.

urmishah:
Yes, this is a typo. Adding the word conveys the original intention better.

samcrang:
@issyl0 after a chat we've tweaked the wording on this to make things (hopefully) clearer.

Thanks again for pointing this out!

PR #113, opened 2018-02-12
[#154518930] Docker registry offline

jonathanglassman:
## What

Added content to manage user expectations when the docker registry that supports their app is offline. Part of story https://www.pivotaltracker.com/story/show/154518930.

## How to review

1. Check that the content covers the story requirements.

## Who can review

Anyone except Jon Glassman or Sam Crang

### DIFF
--- a/source/documentation/deploying_apps/deploying_docker_image.md
+++ b/source/documentation/deploying_apps/deploying_docker_image.md
@@ -28,6 +28,14 @@ There are a few more specifics about Docker image support in Cloud Foundry:
 * Privileged container mode is not supported. Features depending on this will not work.
 * Only registries that use Docker Registry API V2 are supported.
 
+### Offline Docker registry
+
+Deploying, scaling, restarting or restaging an app using a Docker image are all dependent on the Docker registry being available, because these actions require pulling the image from the registry.
+
+Note that Cloud Foundry can restart your application due to circumstances outside of your direct control, such as platform scaling or deployments, or failure of the underlying hardware platform.
+
+If Cloud Foundry attempts to pull the image from the registry when the registry is not available, your application instance will fail to start, and you will receive a `Failed to create container` message. Once the Docker registry is available again, you will be able to re-deploy, restart or restage your app as required.
+
 ### Experimental feature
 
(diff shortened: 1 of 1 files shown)


source/documentation/deploying_apps/deploying_docker_image.md
```diff
@@ -28,6 +28,14 @@ There are a few more specifics about Docker image support in Cloud Foundry:
 * Privileged container mode is not supported. Features depending on this will not work.
 * Only registries that use Docker Registry API V2 are supported.
 
+### Offline Docker registry
+
+An app deployed using a Docker image depends on the availability of the Docker registry that serves that image. 
+
+If Cloud Foundry schedules an instance of your app whilst the registry is offline (for example during a restart or a scale up), Cloud Foundry will not be able to pull the image from the registry. Restarts are normal part of PaaS operation, and can be triggered by events outside of your direct control.
```
samcrang: I'm not sure what "If Cloud Foundry schedules an instance of your app" means?

jonathanglassman: I took the phrasing from the investigation spike summary written by @chrisfarms; Chris do you have any steer on what this means? I thought it was something to do with scheduling an instance of an app for deployment as part of a scale or a restart; am I completely getting that wrong?

paroxp: For instance, CF may find a need of redeploying your application for you, may this bee maintenance of the boxes the application lives in or if application is behaving incorrectly. When that need appears, CF would attempt to download and install Docker Image as it would when user's doing a manual push.

dcarley: If we describe in the opening paragraph that scaling/restarting/restaging are all dependent on registry availability then perhaps this paragraph could focus on platform restarts, something like:

> Cloud Foundry will also restart your application in circumstances outside of your direct control such as scaling, deployments, or underlying hardware failure of the platform itself. If the registry is not available when this happens then your application instance will fail to start.

dcarley: If we describe in the opening paragraph that scaling/restarting/restaging are all dependent on registry availability then perhaps this paragraph could focus on platform restarts, something like:

> Cloud Foundry will also restart your application in circumstances outside of your direct control such as scaling, deployments, or underlying hardware failure of the platform itself. If the registry is not available when this happens then your application instance will fail to start.

(reposting because I deleted it because computers)

PR #118, opened 2018-02-28
[#154331970] Add shared responsibility model section

keymon:
https://www.pivotaltracker.com/story/show/154331970

What?
----

We want to document the responsibility model for the users using Docker
on CF vs. using buildpacks. We can write a document inspired on
https://aws.amazon.com/compliance/shared-responsibility-model/,
including a table or image that would list the different layers
of the application, runtime and platform.

We also add links when explaining buildpacks and docker images.

How to review?
--------------

 - Proof reading.
 - Preview it by running the server locally. See Readme for details.

Who?
----
Anyone but @keymon or @jonathanglassman

### DIFF
--- a/source/documentation/deploying_apps/buildpacks.md
+++ b/source/documentation/deploying_apps/buildpacks.md
@@ -12,7 +12,7 @@ Cloud Foundry uses buildpacks to provide runtime and framework support for appli
 
 We recommend using the standard buildpacks to maximise the support you will receive from GOV.UK PaaS. Using a custom buildpack or Docker image will mean additional setup and maintenance costs.
 
-Your responsibilities change depending on whether you use a standard buildpack, custom buildpack or a docker image. Refer to the [responsibility model guidance](/#responsibility-model) for further information.
+Your responsibilities change depending on whether you use a standard buildpack, custom buildpack or a Docker image. Refer to the [responsibility model guidance](/#responsibility-model) for further information.
 
 ### How to use custom buildpacks
 
--- a/source/documentation/deploying_apps/buildpacks.md
+++ b/source/documentation/deploying_apps/buildpacks.md
@@ -12,7 +12,7 @@ Cloud Foundry uses buildpacks to provide runtime and framework support for appli
 
 We recommend using the standard buildpacks to maximise the support you will receive from GOV.UK PaaS. Using a custom buildpack or Docker image will mean additional setup and maintenance costs.
 
-Refer to the [Guidance](https://docs.cloud.service.gov.uk/#guidance-buildpacks) section for further information on the different options available.
+Your responsibilities change depending on whether you use a standard buildpack, custom buildpack or a docker image. Refer to the [responsibility model guidance](/#responsibility-model) for further information.
 
 ### How to use custom buildpacks
 
--- a/source/documentation/deploying_apps/deploying_docker_image.md
+++ b/source/documentation/deploying_apps/deploying_docker_image.md
@@ -28,7 +28,7 @@ There are a few more specifics about Docker image support in Cloud Foundry:
 * Privileged container mode is not supported. Features depending on this will not work.
 * Only registries that use Docker Registry API V2 are supported.
 
-Your responsibilities change if you use a docker image instead of a buildpack. Refer to the [responsibility model guidance](/#responsibility-model) for further information.
+Your responsibilities change if you use a Docker image instead of a buildpack. Refer to the [responsibility model guidance](/#responsibility-model) for further information.
 
 ### Docker Registry Availability Requirements
 
--- a/source/documentation/deploying_apps/deploying_docker_image.md
+++ b/source/documentation/deploying_apps/deploying_docker_image.md
@@ -28,6 +28,8 @@ There are a few more specifics about Docker image support in Cloud Foundry:
 * Privileged container mode is not supported. Features depending on this will not work.
 * Only registries that use Docker Registry API V2 are supported.
 
+Your responsibilities change if you use a docker image instead of a buildpack. Refer to the [responsibility model guidance](/#responsibility-model) for further information.
+
 ### Docker Registry Availability Requirements
 
 Deploying, scaling, restarting or restaging an app using a Docker image are all dependent on the Docker registry being available, because these actions require pulling the image from the registry.
--- a/source/documentation/guidance/buildpacks.md
+++ b/source/documentation/guidance/buildpacks.md
@@ -45,10 +45,11 @@ Your responsibilities differ depending on whether you use a [standard buildpack]
       </tr>
       <tr>
         <td style="vertical-align: middle; text-align: left">
-          App admin
-          <br> - ensure app instances are up and running
+          App instance lifecycle
+          <br> - ensure app is running
+          <br> - ensure high availability
           <br> - ensure provisioning capacity
-          <br> - stream application logs
+          <br> - stream application logs and metrics
         </td>
         <td style="vertical-align: middle; text-align: center; background-color: #2b8cc4">PaaS</td>
         <td style="vertical-align: middle; text-align: center; background-color: #2b8cc4">PaaS</td>
@@ -66,7 +67,7 @@ Your responsibilities differ depending on whether you use a [standard buildpack]
       </tr>
       <tr>
         <td style="vertical-align: middle; text-align: left">
-          Physical hardware
+          Underlying infrastructure
           <br>- resolve hardware failures
           <br>- encrypt network data traffic
         </td>
--- a/source/documentation/guidance/buildpacks.md
+++ b/source/documentation/guidance/buildpacks.md
@@ -1,44 +1,120 @@
 # Guidance
 
-## Buildpacks
+## Responsibility model
+
+Your responsibilities differ depending on whether you use a [standard buildpack](/#buildpacks), [custom buildpack](/#how-to-use-custom-buildpacks) or [Docker image](/#deploy-a-docker-image-experimental) to deploy your app.
+
+<div class="table-container">
+	<table>
+		<tbody>
+			<tr>
+				<th style="vertical-align: middle; text-align: left">Responsibility</th>
+				<th style="vertical-align: middle; text-align: center">Standard Buildpack</th>
+				<th style="vertical-align: middle; text-align: center">Custom Buildpack</th>
+				<th style="vertical-align: middle; text-align: center">Docker Images</th>
+			</tr>
+			<tr>
+				<td style="vertical-align: middle; text-align: left">
+					Your app
+					<br> - apply security updates to app dependencies
+					<br> - run <a href="https://docs.cloud.service.gov.uk/#penetration-testing">pen tests</a>
+				</td>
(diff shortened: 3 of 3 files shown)


source/documentation/deploying_apps/buildpacks.md
```diff
@@ -12,7 +12,7 @@ Cloud Foundry uses buildpacks to provide runtime and framework support for appli
 
 We recommend using the standard buildpacks to maximise the support you will receive from GOV.UK PaaS. Using a custom buildpack or Docker image will mean additional setup and maintenance costs.
 
-Refer to the [Guidance](https://docs.cloud.service.gov.uk/#guidance-buildpacks) section for further information on the different options available.
+Your responsibilities change depending on whether you use a standard buildpack, custom buildpack or a docker image. Refer to the [responsibility model guidance](/#responsibility-model) for further information.
```
bandesz: capitalisation: `Docker`

source/documentation/deploying_apps/deploying_docker_image.md
```diff
@@ -28,6 +28,8 @@ There are a few more specifics about Docker image support in Cloud Foundry:
 * Privileged container mode is not supported. Features depending on this will not work.
 * Only registries that use Docker Registry API V2 are supported.
 
+Your responsibilities change if you use a docker image instead of a buildpack. Refer to the [responsibility model guidance](/#responsibility-model) for further information.
```
bandesz: capitalisation: `Docker`

source/documentation/guidance/buildpacks.md
```diff
@@ -1,44 +1,120 @@
...
+				</td>
+				<td style="vertical-align: middle; text-align: center; background-color: #ffbf47">Yours</td>
+				<td style="vertical-align: middle; text-align: center; background-color: #ffbf47">Yours</td>
+				<td style="vertical-align: middle; text-align: center; background-color: #ffbf47">Yours</td>
+			</tr>
+			<tr>
+				<td style="vertical-align: middle; text-align: left">
+					Language / runtime
+					<br> - update language and runtime
+          <br> - provide consistent app build process
```
bandesz: use tabs for indentation (or just make it consistent)

source/documentation/guidance/buildpacks.md
```diff
@@ -1,44 +1,120 @@
...
+					<br> - monitor and patch runtime vulnerabilities
```
bandesz: To keep it consistent: - apply security patches

source/documentation/guidance/buildpacks.md
```diff
@@ -1,44 +1,120 @@
...
+					Base operating system
+					<br> - update OS libraries
+				</td>
+				<td style="vertical-align: middle; text-align: center; background-color: #2b8cc4">PaaS</td>
+				<td style="vertical-align: middle; text-align: center; background-color: #2b8cc4">PaaS</td>
+				<td style="vertical-align: middle; text-align: center; background-color: #ffbf47">Yours</td>
+			</tr>
+			<tr>
+				<td style="vertical-align: middle; text-align: left">
+					App admin
```
bandesz: `App admin` sounds strange and the tenants may not understand what we mean by that. Use `Application platform` instead.
Fort the list I suggest:
 - ensure high availability
 - ensure provisioning capacity
 - monitoring

I'm not sure `stream application logs` has any information or value here.

source/documentation/guidance/buildpacks.md
```diff
@@ -1,44 +1,120 @@
...
 
-#### Standard buildpack
-You are only responsible for the app code and its dependencies.
+If you're relatively new to cloud or have a small development team, we recommend using the standard buildpacks to maximise the support you will receive from GOV.UK PaaS.
 
-#### Custom buildpacks
+### Custom buildpacks
 
-You are responsible for managing this custom buildpack as well as the app code and its dependencies. This applies both in setting up the app, and running it in production.
+You are responsible for managing this custom buildpack's language and runtime, as well as the app code and its dependencies. This applies both in setting up the app, and running it in production.
```
bandesz: We are not really talking about a specific custom buildpack, so use `for managing a custom buildpack's`. Also `This applies both _to_`

source/documentation/guidance/buildpacks.md
```diff
@@ -1,44 +1,120 @@
...
 
 When setting up the app:
 
 - Do your own due diligence on finding a maintained and supported custom buildpack so that you have a community to call upon in the case of problems
-- Test deploying with the buildpack early in your build process, not just before go-live
+- Test deploy with the buildpack early in your build process, not just before go-live
```
bandesz: I think `deploying` was better so we don't put two verbs after each other.

source/documentation/guidance/buildpacks.md
```diff
@@ -1,44 +1,120 @@
...
 
 When running the app in production:
 
 - Ensure you have sufficient time to fix issues yourself as using them for deadline-driven development may be risky
 - Regularly maintain your buildpack to update your runtime and dependencies as new security vulnerabilities are discovered and fixed
-- We may not be able to offer support for P1 or out-of-hours incidents; standard SLAs will not apply
+- We may not be able to offer support for P1 or out-of-hours incidents and standard SLAs will not apply
+
+Due to these increased responsibilities, we recommend that, if you must use a custom buildpack, you use a community supported one that is maintained and updated by that community. We do not recommend that you create your own buildpack.
```
bandesz: this is mostly duplicated as we already mentioned this under `When setting up the app`

source/documentation/guidance/buildpacks.md
```diff
@@ -1,44 +1,120 @@
...
 
-#### Docker images
+Contact us at [gov-uk-paas-support@digital.cabinet-office.gov.uk](mailto:gov-uk-paas-support@digital.cabinet-office.gov.uk) if you are interested in this feature. If a custom buildpack is requested by multiple tenants, we will look to improve our support in this area.
```
bandesz: `we will look into improving` sounds better IMO.

source/documentation/guidance/buildpacks.md
```diff
@@ -1,44 +1,120 @@
...
+
+### Docker images
```
bandesz: It seems strange that we detailed standard and custom buildpacks here, but Docker is only a short sentence and a link. Do you think we should move some of the stuff from the Docker section to here?

source/documentation/guidance/buildpacks.md
```diff
@@ -1,44 +1,120 @@
...
+				</td>
+				<td style="vertical-align: middle; text-align: center; background-color: #2b8cc4">PaaS</td>
+				<td style="vertical-align: middle; text-align: center; background-color: #2b8cc4">PaaS</td>
+				<td style="vertical-align: middle; text-align: center; background-color: #2b8cc4">PaaS</td>
+			</tr>
+			<tr>
+				<td style="vertical-align: middle; text-align: left">
+					Physical hardware
+					<br>- resolve hardware failures
+					<br>- encrypt network data traffic
```
bandesz: This sounds like we use hardware encryption :) I think we can move `encrypt network traffic` to the Application platform.

source/documentation/deploying_apps/buildpacks.md
keymon: This is interesting, I recall some discussions that "Docker" with upper case refers to the company and the product, then use "docker image"  "docker network", etc to refer to the tech stack.

But I am happy to change it.

source/documentation/guidance/buildpacks.md
keymon: I wouldn't put *monitoring* because we do not do any monitoring at all.

keymon: Instead of *Application platform* , what about *Application lifecycle*

> Application lifecycle
>  * ensure high availability
>  * ensure provisioning capacity
>  * stream application logs and metrics

source/documentation/guidance/buildpacks.md
keymon: The section is *Custom buildpacks*

I do not understand the other comment `This applies both _to_`, I guess you refer to grammar. @jonathanglassman any comment on this?

source/documentation/guidance/buildpacks.md
keymon: /cc @jonathanglassman

source/documentation/guidance/buildpacks.md
bandesz: This is the Cloud Foundry layer itself, so I would call it platform and not lifecycle. And we do monitoring there, e.g. if your app dies then we restart it automatically.

Monitoring the apps is a responsibility.

source/documentation/guidance/buildpacks.md
bandesz: My first comment is about replacing `this` with `a`.
The second one is to replace `in` with `to`.

jonathanglassman: Agree with @bandesz. Will change to `You are responsible for managing a custom buildpack's language and runtime, as well as the app code and its dependencies. This applies both to setting up the app, and running it in production.`

source/documentation/guidance/buildpacks.md
jonathanglassman: I would go with `Test deploy` as it is more active than `Test deploying`.

source/documentation/guidance/buildpacks.md
jonathanglassman: True. How about the bullet point reads `Use a buildpack that has been created and is maintained by a community (rather than creating your own buildpack), so that you can ask that community for support with any issues`, and we remove the duplicate content?

source/documentation/guidance/buildpacks.md
jonathanglassman: Are we trying not to commit to anything on this?

source/documentation/guidance/buildpacks.md
jonathanglassman: I agree but I think that should be looked at as part of an overall change of the Docker section as it is not part of this story. I think the Docker section overall needs some work, so I recommend we draw a line under this and look at this later. What do you think?

source/documentation/guidance/buildpacks.md
bandesz: ok

source/documentation/guidance/buildpacks.md
bandesz: Sorry, but this expression `look to improve` sounds strange to me and I didn't find any examples that people use it like this. But that doesn't mean it's not legit :)

source/documentation/guidance/buildpacks.md
bandesz: I'm fine with that.

source/documentation/guidance/buildpacks.md
jonathanglassman: ok, I don't feel strongly about it, I'll put it back

source/documentation/guidance/buildpacks.md
jonathanglassman: 👍

PR #120, opened 2018-03-13
[#155439160] Connect to Redis through Conduit

camelpunch:
## What

Describe how a [PaaS Elasticache Broker](https://github.com/alphagov/paas-elasticache-broker) user can connect to their service instance from their local machine using [Conduit](https://github.com/alphagov/paas-cf-conduit). 

## How to review

1. Check the content addresses the story

## Who can review

Anyone except @jonathanglassman and @camelpunch.

### DIFF
--- a/source/documentation/deploying_services/index.md
+++ b/source/documentation/deploying_services/index.md
@@ -1065,6 +1065,7 @@ You must:
     options=NO_SSLv2
     options=NO_SSLv3
     foreground=yes
+    pid=/tmp/stunnel.pid
     [redis]
     client = yes
     accept = 127.0.0.1:6379
@@ -1080,7 +1081,7 @@ You must:
 1. Start stunnel by running:
 
     ```
-    sudo stunnel CONFIGURATION_FILE
+    stunnel CONFIGURATION_FILE
     ```
 
 1. Connect to your Redis service instance by running:
--- a/source/documentation/deploying_services/index.md
+++ b/source/documentation/deploying_services/index.md
@@ -191,7 +191,7 @@ Run `cf conduit --help` for more options, and refer to the [Conduit readme file]
 
 ### Import and export bulk data to and from a PostgreSQL database
 
-#### Prerequisites 
+#### Prerequisites
 
 You must:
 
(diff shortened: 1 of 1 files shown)


source/documentation/deploying_services/index.md
```diff
@@ -1008,6 +1005,102 @@ You must bind your app to the Redis service to be able to access the cache from
 
     If your app writes service connection errors to `STDOUT` or `STDERR`, you can view recent errors with ``cf logs APPNAME --recent``. See the section on [Logs](#logs) for details.
 
+### Connect to a Redis service instance from your local machine
+
+You must use the [Conduit](#conduit) plugin and stunnel tool to connect your local machine to your Redis service instance.
```
samcrang: Is it worth linking to the section on installing `stunnel`?

source/documentation/deploying_services/index.md
```diff
@@ -1008,6 +1005,102 @@ You must bind your app to the Redis service to be able to access the cache from
...
+    where:
+    - `127.0.0.1:6379` is the default Redis host and port that stunnel will open (you will need to use a different port if you are already running a local Redis service instance)
+    - `127.0.0.1:7081` is the default Conduit host and port outputted by the `cf conduit SERVICE_NAME` command
+
+#### Connect your local machine to your Redis service instance
+
+1. Start stunnel by running:
+
+    ```
+    sudo stunnel CONFIGURATION_FILE
```
samcrang: Why does it require `sudo`?

camelpunch: Because I blindly copied the AWS docs, which were blindly copied from some user. If you set a pidfile, you don't need sudo I think. Just testing.

source/documentation/deploying_services/index.md
```diff
@@ -1008,6 +1005,102 @@ You must bind your app to the Redis service to be able to access the cache from
...
+    host: 127.0.0.1
+    port: 7081
+    username:
+    password: REDIS_PASSWORD
+    db: DATABASE
+    ```
+
+    where `REDIS_PASSWORD` is the password given to you by Conduit.
+
+#### Install and configure stunnel
```
dcarley: A TLS proxy would be a nice feature to add to conduit in the future 🤔

alext: Indeed, we talked about this, but decided to document the current approach in the meantime.

source/documentation/deploying_services/index.md
alext: The word "must" here seems a bit strong. There's no requirement to use conduit to achieve this. It's probably the easiest way to do it, but there are other approaches such as manually creating ssh tunnels.

I also wonder whether we need to call out why stunnel is necessary (ie the redis-cli doesn't have TLS support). If using a different client that does have TLS support, stunnel wouldn't be necessary.

source/documentation/deploying_services/index.md
camelpunch: New commit added.

PR #124, opened 2018-04-05
Upgrade to latest tech docs

tijmenb:
## What

This upgrades the tech docs to the latest version, which uses the [govuk_tech_docs gem](https://github.com/alphagov/tech-docs-gem) to pull in CSS, JS, fonts, images and layouts. This means that a lot of files can be removed and only customised code and configuration remains. You can read more about it in https://github.com/alphagov/tech-docs-template/pull/144.

I'm updating projects that haven't upgraded yet as part of the GOV.UK firebreak, in the hope that if we are all on the same version it's more likely that sites will keep up to date and contribute back to the main project.

This PR is almost the same as https://github.com/alphagov/reliability-engineering/pull/10, https://github.com/alphagov/gds-way/pull/132 and https://github.com/alphagov/registers-tech-docs/pull/34.

## How to review

There should be no changes in the layout or behaviour after this upgrade. I've checked this by comparing https://docs.cloud.service.gov.uk to the new site. I didn't spot any differences.

## Who can review

Anyone but me.

---

Part of the Q4 GOV.UK 🔥 firebreak: https://trello.com/c/QbspvRc2


LeePorte:
Thanks for the PR @tijmenb

PR #136, opened 2018-05-10
Data official

jonathanglassman:
What
----

Added data security classification section to the Deploying apps and backing services sections of the tech docs, stating that the PaaS can store data up to "official".

How to review
-------------

1. Check that it covers https://www.pivotaltracker.com/story/show/156810789

Who can review
--------------

Anyone except Lee Porte and Jon Glassman

### DIFF
--- a/source/documentation/deploying_apps/index.md
+++ b/source/documentation/deploying_apps/index.md
@@ -32,6 +32,14 @@ For a production app, you should read the [production checklist](/#production-ch
 * Instances will be restarted if they [exceed memory limits](/#quotas).
 * Your application should write all its log messages to `STDOUT`/`STDERR`, rather than a log file.
 
+## Data security classification
+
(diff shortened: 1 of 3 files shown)


venusbb:
In ```https://github.com/alphagov/paas-tech-docs/blob/master/source/documentation/overview/benefits.md``` we use **OFFICIAL**, the new added paragraphs, we use "official". We should make this consistent.

venusbb:
In https://www.cloud.service.gov.uk/features we also has this information, but Official is used. It would be good to get the quotation consistent.

jonathanglassman:
Thanks Venus. The service manual uses ‘official’ so I'll change this to bring it in line with the service manual. This will also be consistent with https://www.cloud.service.gov.uk/ia and https://www.cloud.service.gov.uk/security.

PR #154, opened 2018-07-19
Update govuk-tech-docs gem

jonathanglassman:
What
----
Updated the gemfile.lock to use the latest gems. This is primarily to implement the new search functionality on the tech docs.

How to review
-------------

1. Check that the search function works.

Who can review
--------------

Anyone except Jon Glassman


samcrang:
It looks like (for reasons I don't understand) that we have to use a [GDS fork of `middleman-search`](https://github.com/alphagov/tech-docs-template/pull/156/commits/c7bf3d53b6a1e636ca2a50205a1ecaa80ae28ae6#diff-c44a3f126688f4b87c6ef2491ab60a07R15). As well as [enabling search in the config](https://github.com/alphagov/tech-docs-gem/blob/aee8e589cebd5293cb58eaa2008bcde73a29b730/example/config/tech-docs.yml#L36).

I'll update the commits.

PR #157, opened 2018-08-03
Drop syslog_msg if it's the same as message field

sakisv:
What
----

I noticed last week in Kibana that several logs contained the same data in two fields,`syslog_msg` and `message` which was affecting the space required on logit. The effect would be even bigger if taken together with the fact that new versions of elasticsearch index a string field both as a `text` and as `keyword`. ([docs](https://www.elastic.co/blog/strings-are-dead-long-live-strings))

This commit updates the logstash filter so that when all parsing is finished, it will compare the two fields and drop the `syslog_msg` if they are the same.

How to review
-------------

1. Identify that this problem exists on existing stacks
2. Apply the change on that stack's logstash
3. Compare the differences

Who can review
--------------

Not me.

### DIFF
--- a/source/documentation/monitoring_apps/logs.md
+++ b/source/documentation/monitoring_apps/logs.md
@@ -112,6 +112,12 @@ filter {
             target => "[access][user_agent]"
         }
     }
+
+    if [syslog_msg] == [message] {
+        mutate {
+            remove_field => ["syslog_msg"]
+        }
+    }
 }
 ```
 


keymon:
Hi, thank you for your PR. 

You are absolutely right. But I wonder, instead of adding that rule at the end, wouldn't it be better to add the line 

```
            remove_field => ["syslog_msg"]
```

just [after line 69, where the field is copied and duplicated?](https://github.com/alphagov/paas-tech-docs/pull/157/files#diff-2d06baeb819bb70891f8facf55df6c64L69):

```diff
--- a/source/documentation/monitoring_apps/logs.md
+++ b/source/documentation/monitoring_apps/logs.md
@@ -67,6 +67,7 @@ filter {
         mutate {
             replace => [ "source_host", "%{syslog_host}" ]
             replace => [ "message", "%{syslog_msg}" ]
+            remove_field => ["syslog_msg"]
         }
     }
```

keymon:
I tried using `rename` instead but it did not work


```diff
--- a/source/documentation/monitoring_apps/logs.md
+++ b/source/documentation/monitoring_apps/logs.md
@@ -65,8 +65,8 @@ filter {
     if !("_syslogparsefailure" in [tags]) {
         # if we successfully parsed syslog, replace the message and source_host fi
         mutate {
-            replace => [ "source_host", "%{syslog_host}" ]
-            replace => [ "message", "%{syslog_msg}" ]
+            rename => [ "source_host", "%{syslog_host}" ]
+            rename => [ "message", "%{syslog_msg}" ]
         }
     }
```

sakisv:
The `syslog_msg` field is [used in line 84](https://github.com/alphagov/paas-tech-docs/pull/157/files#diff-2d06baeb819bb70891f8facf55df6c64L84) when parsing the router logs.

I guess we could get the entire section between lines 65-71 and put it in the end of the processing?

keymon:
Ops, you are right.

So something more like this: 

```diff
    if !("_syslogparsefailure" in [tags]) {
        # if we successfully parsed syslog, replace the message and source_host fields
        mutate {
            rename => { "syslog_host" => "source_host" }
            rename => { "syslog_msg" => "message" }
        }
    }
```

keymon:
That seems to work fine, and the `access.*` fields are populated.

keymon:
@sakisv do you mind change the PR and I will be happy to merge this :smile:

sakisv:
Done!

PR #158, opened 2018-08-07
document unencrypted to encrypted postgres migration

gidsg:
What
----

Added documentation for migrating from unencrypted to encrypted postgres service plan, based on recent experience doing this on Registers

How to review
-------------
Following the steps you should be able to migrate from an unencrypted to an encrypted plan. Note: I have not covered handling database downtime or a read-only period after taking a snapshot as I considered those out of scope and they may vary depending on your application.

Who can review
--------------
@46bit or anyone else in PAAS team.


jonathanglassman:
Actioned as part of https://github.com/alphagov/paas-tech-docs/pull/164

PR #159, opened 2018-08-08
Elasticsearch

jonathanglassman:
What
----

Expanded Elasticsearch section within Deploying Backing Services section of tech docs. This is an MVP and has spun off multiple other stories for investigation.

How to review
-------------

1. Check that the content fulfils story https://www.pivotaltracker.com/story/show/154638140

Who can review
--------------

Anyone except Jon Glassman or Jonathan Matthews.

### DIFF
--- a/source/documentation/deploying_services/index.md
+++ b/source/documentation/deploying_services/index.md
@@ -1171,19 +1171,216 @@ Refer to the [Amazon ElastiCache for Redis page](https://aws.amazon.com/elastica
 
 ## Elasticsearch
 
+## Elasticsearch
+
+___Does this need to be changed?___
+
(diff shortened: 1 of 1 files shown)


jpluscplusm:
@keymon Whilst Aiven take backups, they're not exposed via the broker as far as I can see. Also, the nature of the data that a tenant "should" be holding in ES is the kind which can be recreated from their primary datastore relatively easily. Taking these together was why @jonathanglassman and I chose not to keep a backup/restore section in these docs.

As the broker stands right now, the only thing we could document is that (1 + 2 hours ago) backups exist, but it's a manual operation to use them in any way. Is that worth doing, given the non-primary-datastore nature of the service?

keymon:
> Taking these together is why @jonathanglassman and I chose not to keep a backup/restore section in these docs.

My only concern is that the tenants might ask or wonder about it. @jonathanglassman let me know what you think, I am happy to merge this as it is.

jonathanglassman:
After discussion with @keymon and @jpluscplusm have agreed to state:

Backups are not currently supported. As a non-primary datastore, we recommend that transferring your data from your primary datastore to ES is something that your app is capable of doing on demand.

jonathanglassman:
@keymon see latest commit - ok to approve?

PR #185, opened 2018-11-15
Upgrade rack to 2.0.6

alext:
## What

To fix CVE-2018-16470 [1]. This isn't a significant thing for us as
this is a static site generator, so rack isn't running in production,
but it's worth updating to avoid the noise etc.

[1]https://groups.google.com/forum/#!msg/rubyonrails-security/U_x-YkfuVTg/xhvYAmp6AAAJ

How to review
-------------

Code review should be enough

Who can review
--------------

Not me.


richardTowers:
Code review looks fine (two bits changed). Tested the build locally and it still works 👍

PR #188, opened 2018-12-03
document how to set JRE version using env var

gidsg:
What
----
The previous text had some information under the heading *Specify a Java Version* about checking the buildpack version which is irrelevant to specifying the Java version!
  
Judging by my own experience and others on Slack this information was not helpful and is arguably harmful in sending people further away from the expected answer.

I have changed the text to reflect the required steps to set the JRE version.

How to review
-------------

run locally and go to: `/deploying_apps.html#specify-a-java-version`

Who can review
--------------
anyone but me. CC @richardTowers

### DIFF
--- a/source/documentation/deploying_apps/deploying_java.md
+++ b/source/documentation/deploying_apps/deploying_java.md
@@ -112,9 +112,7 @@ You can specify the version of the Java Runtime Environment (JRE) by setting the
   env:
     JBP_CONFIG_OPEN_JDK_JRE: '{ jre: { version: 11.+ } }'
  ```
-The `+` in the above example acts as a wildcard which will use the latest point release available.
-
-To see if a JRE version is available you can check this [version list](http://download.pivotal.io.s3.amazonaws.com/openjdk/trusty/x86_64/index.yml) [external link]. 
+The `+` in the above example acts as a wildcard which will use the latest point release available. 
 
 ### Deploying other JVM-based applications
 
--- a/source/documentation/deploying_apps/deploying_java.md
+++ b/source/documentation/deploying_apps/deploying_java.md
@@ -112,7 +112,7 @@ You can specify the version of the Java Runtime Environment (JRE) by setting the
   env:
     JBP_CONFIG_OPEN_JDK_JRE: '{ jre: { version: 11.+ } }'
  ```
-The `+` in the above example acts as a wildcard which will use the latest point release available. 
+The `+` in this example means it will use the latest point release available.
 
 ### Deploying other JVM-based applications
 
--- a/source/documentation/deploying_apps/deploying_java.md
+++ b/source/documentation/deploying_apps/deploying_java.md
@@ -106,7 +106,7 @@ For more configuration options see [Tomcat Container](https://github.com/cloudfo
 
 ### Specify a Java version
 
-You can specify the version of the Java Runtime Environment (JRE) by setting the environment variable `JBP_CONFIG_OPEN_JDK_JRE`. For example to use JRE 11 you could add the following to your application's `manifest.yml`:
+You can specify the version of the Java Runtime Environment (JRE) by setting the environment variable `JBP_CONFIG_OPEN_JDK_JRE` in the application's `manifest.yml`. For example, to use JRE 11:
 
 ```yaml
   env:
--- a/source/documentation/deploying_apps/deploying_java.md
+++ b/source/documentation/deploying_apps/deploying_java.md
@@ -106,11 +106,15 @@ For more configuration options see [Tomcat Container](https://github.com/cloudfo
 
 ### Specify a Java version
 
-You should tell Cloud Foundry which version of Java your app uses in the Java buildpack.
+You can specify the version of the Java Runtime Environment (JRE) by setting the environment variable `JBP_CONFIG_OPEN_JDK_JRE`. For example to use JRE 11 you could add the following to your application's `manifest.yml`:
 
-Refer to the documentation on [buildpack language version updates](deploying_apps.html#buildpack-language-version-updates) for more information.
+```yaml
+  env:
+    JBP_CONFIG_OPEN_JDK_JRE: '{ jre: { version: 11.+ } }'
+ ```
+The `+` in the above example acts as a wildcard which will use the latest point release available.
 
-Refer to the Cloud Foundry OpenJDK JRE documentation for more information on [the Java buildpack](https://github.com/cloudfoundry/java-buildpack/blob/master/docs/jre-open_jdk_jre.md) [external link].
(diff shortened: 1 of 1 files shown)


source/documentation/deploying_apps/deploying_java.md
```diff
@@ -106,11 +106,15 @@ For more configuration options see [Tomcat Container](https://github.com/cloudfo
 
 ### Specify a Java version
 
-You should tell Cloud Foundry which version of Java your app uses in the Java buildpack.
+You can specify the version of the Java Runtime Environment (JRE) by setting the environment variable `JBP_CONFIG_OPEN_JDK_JRE`. For example to use JRE 11 you could add the following to your application's `manifest.yml`:
```
jonathanglassman: Is this the only way to do this?

if so, suggest the following: "Specify the version of the Java Runtime Environment (JRE) by setting the environment variable `JBP_CONFIG_OPEN_JDK_JRE` in the app's `manifest.yml`. For example, to use JRE 11:

source/documentation/deploying_apps/deploying_java.md
```diff
@@ -106,11 +106,15 @@ For more configuration options see [Tomcat Container](https://github.com/cloudfo
...
 
-Refer to the documentation on [buildpack language version updates](deploying_apps.html#buildpack-language-version-updates) for more information.
+```yaml
+  env:
+    JBP_CONFIG_OPEN_JDK_JRE: '{ jre: { version: 11.+ } }'
+ ```
+The `+` in the above example acts as a wildcard which will use the latest point release available.
```
jonathanglassman: Suggest we say `version: 11.x` and remove line 115 as this should be known to the user.

source/documentation/deploying_apps/deploying_java.md
richardTowers: You could use `cf set-env ...` instead of the manifest if you wanted to, but in this case the manifest is a much more sane approach I think. Might be ok just to document the manifest as the only way, even though there's technically another approach.

Could link out to [#environment-variables](https://docs.cloud.service.gov.uk/deploying_apps.html#environment-variables) though.

source/documentation/deploying_apps/deploying_java.md
richardTowers: >  as this should be known to the user

I don't think it necessarily will be - this confused me when I was trying to get Java 11 running, I couldn't work out whether to use `+` or `x` or `*` or what.

richardTowers:
This is the bit in the upstream documentation that mentions setting this environment variable to set the java version: https://github.com/cloudfoundry/java-buildpack/blob/aa723671773b05baab0858334f7f9340bff15eb9/README.md#configuration-and-extension

source/documentation/deploying_apps/deploying_java.md
```diff
@@ -106,11 +106,15 @@ For more configuration options see [Tomcat Container](https://github.com/cloudfo
...
 
-Refer to the Cloud Foundry OpenJDK JRE documentation for more information on [the Java buildpack](https://github.com/cloudfoundry/java-buildpack/blob/master/docs/jre-open_jdk_jre.md) [external link].
+To see if a JRE version is available you can check this [version list](http://download.pivotal.io.s3.amazonaws.com/openjdk/trusty/x86_64/index.yml) [external link].
```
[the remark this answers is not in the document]
gidsg: not sure there is a better link, this is what's described and linked to from: https://github.com/cloudfoundry/java-buildpack/blob/master/docs/extending-repositories.md#repository-structure unless we think this part is extraneous then maybe just delete it?

source/documentation/deploying_apps/deploying_java.md
gidsg: I took this example style from the tomcat one in https://github.com/alphagov/paas-tech-docs/blob/e2958ca162277f99952146b8f00511d56cd9bde0/source/documentation/deploying_apps/deploying_java.md#deploying-a-war-file it does say *for example* so I would say that implies it is not the only approach but I have found it is the most helpful for example when helping people on Slack. Not precious about this if you want to change it though.

source/documentation/deploying_apps/deploying_java.md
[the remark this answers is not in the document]
richardTowers: I think maybe delete it - I think it's a bit jarring to follow a link out to plaintext YAML (I'm also not 100% sure how accurate this list is - is it just an example of what the index should look like, or is it an authoritative source? I can't tell.)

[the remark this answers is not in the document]
gidsg: @richardTowers I believe it is authoritative as per: https://github.com/cloudfoundry/java-buildpack/blob/master/docs/extending-repositories.md#wildcards i.e. it is the default `repository.root` if not overriden. However, if you think it is confusing I will just remove it. I do find it useful to know when a new JRE is available but I guess this is a less common need than setting a value.

source/documentation/deploying_apps/deploying_java.md
jonathanglassman: How about a compromise: "You can specify the version of the Java Runtime Environment (JRE) by setting the environment variable JBP_CONFIG_OPEN_JDK_JRE in the app's manifest.yml. For example, to use JRE 11"?

source/documentation/deploying_apps/deploying_java.md
jonathanglassman: ok

source/documentation/deploying_apps/deploying_java.md
```diff
@@ -106,11 +106,13 @@ For more configuration options see [Tomcat Container](https://github.com/cloudfo
...
+You can specify the version of the Java Runtime Environment (JRE) by setting the environment variable `JBP_CONFIG_OPEN_JDK_JRE` in the application's `manifest.yml`. For example, to use JRE 11:
 
-Refer to the documentation on [buildpack language version updates](deploying_apps.html#buildpack-language-version-updates) for more information.
-
-Refer to the Cloud Foundry OpenJDK JRE documentation for more information on [the Java buildpack](https://github.com/cloudfoundry/java-buildpack/blob/master/docs/jre-open_jdk_jre.md) [external link].
+```yaml
+  env:
+    JBP_CONFIG_OPEN_JDK_JRE: '{ jre: { version: 11.+ } }'
+ ```
+The `+` in the above example acts as a wildcard which will use the latest point release available.
```
jonathanglassman: Looks good. Just a couple of points:

- "wildcard" is a bit colloquial so is not that friendly to non-native speakers. 
- "above" is on the list of words to avoid because it is not friendly to screenreaders

Would recommend changing to "The `+` in this example is the latest point release available."

jonathanglassman: "means it will use" sounds good. So let's combine: "The + in this example means it will use the latest point release available."

PR #192, opened 2019-01-03
[#162793860] Prometheus exporter documentation fixes

bandesz:
What
----

* Fix: the paas-metric-exporter is not displaying any metrics but pushes them to a StatsD-compatible server
* Change the order that users should start the prometheus-exporter app first and then set it up in Prometheus
* Fix type: supposed -> supported

How to review
-------------

Code review. I don't think a 2i review is needed as there is no new content.

Who can review
--------------

Anyone.


jonathanglassman:
LGTM

PR #202, opened 2019-03-08
Enable extensions in PostgreSQL service instances

jonathanglassman:
What
----

- Adds content on enabling and disabling extensions in PostgreSQL service instances.
- Removes unneeded content on extensions whitelist

How to review
-------------

1. Check it fulfils https://www.pivotaltracker.com/story/show/163979461

Who can review
--------------

Anyone except Jon Glassman.

### DIFF
--- a/source/documentation/deploying_services/postgresql.md
+++ b/source/documentation/deploying_services/postgresql.md
@@ -271,9 +271,9 @@ cf update-service SERVICE_NAME -c '{"reboot": true, "force_failover": true}'
 
 When you force a failover, your PostgreSQL database IP address will change. The database's hostname will not change. You must configure your app to close all database connections to the previous IP address after forcing a failover.
 
-### Add or remove extensions for a Postgresql service instance
+### Add or remove extensions for a PostgreSQL service instance
 
-The following extensions are always enabled for Postgresql service instances:
+The following extensions are always enabled for PostgreSQL service instances:
 
 <div style="height:1px;font-size:1px;">&nbsp;</div>
 
@@ -285,7 +285,7 @@ The following extensions are always enabled for Postgresql service instances:
 
 <div style="height:1px;font-size:1px;">&nbsp;</div>
 
-You cannot remove these mandatory extensions from a Postgresql service instance.
+You cannot remove these mandatory extensions from a PostgreSQL service instance.
 
 GOV.UK PaaS supports the following optional extensions, but they are not enabled by default:
 
@@ -303,7 +303,7 @@ You can enable or disable optional extensions for a PostgreSQL service instance 
 - updating an existing service instance
 - restoring a PostgreSQL service instance snapshot
 
-#### Create a new service instance
+#### Enable an optional extension when creating a service instance
 
 You [create a new PostgreSQL service instance](#set-up-a-postgresql-service) by running `cf create-service`. You can enable optional extensions in this new service instance by running:
 
@@ -337,7 +337,9 @@ For example, your PostgreSQL service instance is named `my-pg-service` and has 3
 cf update-service my-pg-service -c '{"enable_extensions": ["pg_stat_statements"], "reboot": true}'
 ```
 
-Rebooting a service instance will cause some service downtime. You should schedule this downtime to minimise service disruption. Refer to the [documentation on rebooting a service instance](https://docs.cloud.service.gov.uk/deploying_services/postgresql/#reboot-a-postgresql-service-instance) for more information.
+Rebooting a service instance will cause some service downtime. The length of this outage depends on your service instance’s complexity and configuration. 
+
+You should schedule this downtime to minimise service disruption. Refer to the [documentation on rebooting a service instance](https://docs.cloud.service.gov.uk/deploying_services/postgresql/#reboot-a-postgresql-service-instance) for more information.
 
 #### Disable an optional extension in an existing service instance
 
@@ -362,9 +364,9 @@ cf update-service my-pg-service -c '{"disable_extensions": ["pg_stat_statements"
 
 Rebooting a service instance will cause some service downtime. You should schedule this downtime to minimise service disruption. Refer to the [documentation on rebooting a service instance](https://docs.cloud.service.gov.uk/deploying_services/postgresql/#reboot-a-postgresql-service-instance) for more information.
 
-#### Add extensions when restoring a service instance
+#### Enable an optional extension when restoring a service instance
 
-When you [restore a PostgreSQL service instance from a snapshot](#restoring-a-postgresql-service-snapshot), that restored service instance will include all extensions that were enabled at time of backup.
+When you [restore a PostgreSQL service instance from a snapshot](#restoring-a-postgresql-service-snapshot), that restored service instance will include all extensions enabled at time of backup.
 
 You can also enable optional extensions when restoring a service instance:
 
@@ -379,6 +381,7 @@ where:
 - `EXTENSION_1...N` are the names of the optional extensions you want to enable
 
 Refer to the documentation on [restoring a PostgreSQL service snapshot](#restoring-a-postgresql-service-snapshot) for more information.
+
 
 ## Remove the service
 
--- a/source/documentation/deploying_services/postgresql.md
+++ b/source/documentation/deploying_services/postgresql.md
@@ -271,6 +271,93 @@ cf update-service SERVICE_NAME -c '{"reboot": true, "force_failover": true}'
 
 When you force a failover, your PostgreSQL database IP address will change. The database's hostname will not change. You must configure your app to close all database connections to the previous IP address after forcing a failover.
 
+### Add or remove extensions for a Postgresql service instance
+
+GOV.UK PaaS supports and enables the following extensions by default:
+
+<div style="height:1px;font-size:1px;">&nbsp;</div>
+
+|Extension|PostgreSQL version|
+|:---|:---|
+|uuid-ossp|9.5, 10|
+|postgis|9.5, 10|
+|citext|10|
+
+<div style="height:1px;font-size:1px;">&nbsp;</div>
+
+GOV.UK PaaS supports the following extensions, but they are not enabled by default:
+
+<div style="height:1px;font-size:1px;">&nbsp;</div>
+
+|Extension|PostgreSQL version|
+|:---|:---|
+|pg_stat_statements|9.5, 10|
+
+<div style="height:1px;font-size:1px;">&nbsp;</div>
+
+You can add or remove extensions for a PostgreSQL service instance by:
+
+- creating a new service instance
+- updating an existing service instance
+- restoring a PostgreSQL service instance snapshot
+
+#### Create a new service instance
+
+You [create a new PostgreSQL service instance](#set-up-a-postgresql-service) by running `cf create-service`. You can enable extensions in this new service instance by running:
+
+```
+cf create-service SERVICE_NAME -c '{"enabled_extensions": ["EXTENSION_1","EXTENSION_2",...,"EXTENSION_N"]}'
+```
+
+where:
+
+- `SERVICE_NAME` is a unique descriptive name for this service instance
+- `EXTENSION_1...N` are the names of the extensions you want to enable
+
+#### Update an existing service instance
+
+You add or remove extensions on an existing PostgreSQL service instance by updating your service with the new list of extensions:
+
+```
+cf update-service SERVICE_NAME -c '{"enabled_extensions": ["EXTENSION_1","EXTENSION_2",...,"EXTENSION_N"]}'
+```
+
+For example, your PostgreSQL service instance is named `my-pg-service` and has 3 extensions enabled, `uuid-ossp`, `postgis` and `citext`. Run the following to remove `citext`:
+
+```
+cf update-service my-pg-service -c '{"enabled_extensions": ["uuid-ossp","postgis"]}'
+```
+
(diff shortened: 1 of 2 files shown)


source/documentation/deploying_services/postgresql.md
```diff
@@ -271,6 +271,93 @@ cf update-service SERVICE_NAME -c '{"reboot": true, "force_failover": true}'
...
+cf update-service SERVICE_NAME -c '{"enabled_extensions": ["EXTENSION_1","EXTENSION_2",...,"EXTENSION_N"]}'
+```
+
+For example, your PostgreSQL service instance is named `my-pg-service` and has 3 extensions enabled, `uuid-ossp`, `postgis` and `citext`. Run the following to remove `citext`:
+
+```
+cf update-service my-pg-service -c '{"enabled_extensions": ["uuid-ossp","postgis"]}'
+```
+
+If you add the `pg_stat_statements` extension to your service instance, this extension will not start until the GOV.UK PaaS team reboots your service instance.
```
AP-Hunt: We (as a team) don't reboot the instance. Part of the update story will be to work out how and when it should be triggered automatically by the broker.

source/documentation/deploying_services/postgresql.md
```diff
@@ -271,6 +271,93 @@ cf update-service SERVICE_NAME -c '{"reboot": true, "force_failover": true}'
...
+
+```
+cf create-service postgres PLAN NEW_SERVICE_NAME -c '{"restore_from_latest_snapshot_of": "GUID", "enabled_extensions": ["EXTENSION_1","EXTENSION_2",...,"EXTENSION_N"]}'
+```
+where:
+
+- `PLAN` is the plan used in the original service instance
+- `NEW_SERVICE_NAME` is a unique, descriptive name for the restored service instance
+- `GUID` is the global unique identifier of the backed up service instance
+- `EXTENSION_1...N` are the names of the extensions you want to enable
```
AP-Hunt: Reading this and the last section, I can see how it might be unclear that these are _additional_ extensions, and not an exclusive list as it usually would be.

source/documentation/deploying_services/postgresql.md
bandesz: There is a parameter called `apply_at_maintenance_window`. I suspect we will use it here too, so if tenants set it to false then the database instance will be rebooted immediately.

Also tenants can reboot the instance manually as we documented here: https://docs.cloud.service.gov.uk/deploying_services/postgresql/#reboot-a-postgresql-service-instance

source/documentation/deploying_services/postgresql.md
```diff
@@ -271,6 +271,93 @@ cf update-service SERVICE_NAME -c '{"reboot": true, "force_failover": true}'
...
+
+This reboot will cause some service downtime. You should schedule this downtime to minimise service disruption.
+
+#### Restore a service instance
+
+If you have a [paid PostgreSQL plan](#paid-plans-postgresql), the Amazon RDS system [backs up your PostgreSQL service data](#postgresql-service-backup) every night. This includes all extensions that were enabled at time of backup.
+
+When you [restore a PostgreSQL service instance from a snapshot](#restoring-a-postgresql-service-snapshot), that restored service instance will by default include these extensions.
+
+If the snapshot contains extensions that GOV.UK PaaS no longer supports, the restored service instance will still include these extensions. However, you cannot create new service instances with unsupported extensions.
```
bandesz: I'm not sure whether we should include this as this is theoretical and we don't know whether it's true or not. I would remove it.

source/documentation/deploying_services/postgresql.md
```diff
@@ -271,6 +271,93 @@ cf update-service SERVICE_NAME -c '{"reboot": true, "force_failover": true}'
...
+
+#### Update an existing service instance
+
+You add or remove extensions on an existing PostgreSQL service instance by updating your service with the new list of extensions:
+
+```
+cf update-service SERVICE_NAME -c '{"enabled_extensions": ["EXTENSION_1","EXTENSION_2",...,"EXTENSION_N"]}'
+```
+
+For example, your PostgreSQL service instance is named `my-pg-service` and has 3 extensions enabled, `uuid-ossp`, `postgis` and `citext`. Run the following to remove `citext`:
```
bandesz: This example is misleading, as we don't plan to allow to disable default plugins. The only plugin that can be enabled/disabled is the `pg_stat_statements` currently.

source/documentation/deploying_services/postgresql.md
```diff
@@ -271,6 +271,93 @@ cf update-service SERVICE_NAME -c '{"reboot": true, "force_failover": true}'
 
 When you force a failover, your PostgreSQL database IP address will change. The database's hostname will not change. You must configure your app to close all database connections to the previous IP address after forcing a failover.
 
+### Add or remove extensions for a Postgresql service instance
+
+GOV.UK PaaS supports and enables the following extensions by default:
```
bandesz: We should say the following extensions are always enabled.

source/documentation/deploying_services/postgresql.md
```diff
@@ -271,6 +271,93 @@ cf update-service SERVICE_NAME -c '{"reboot": true, "force_failover": true}'
...
+```
+
+where:
+
+- `SERVICE_NAME` is a unique descriptive name for this service instance
+- `EXTENSION_1...N` are the names of the extensions you want to enable
+
+#### Update an existing service instance
+
+You add or remove extensions on an existing PostgreSQL service instance by updating your service with the new list of extensions:
```
bandesz: We should leave out removing extensions completely and only add it when we really plan to support removal. (This might depend partly on the outcome of the update story)

source/documentation/deploying_services/postgresql.md
```diff
@@ -271,6 +271,93 @@ cf update-service SERVICE_NAME -c '{"reboot": true, "force_failover": true}'
...
+
+If you have a [paid PostgreSQL plan](#paid-plans-postgresql), the Amazon RDS system [backs up your PostgreSQL service data](#postgresql-service-backup) every night. This includes all extensions that were enabled at time of backup.
+
+When you [restore a PostgreSQL service instance from a snapshot](#restoring-a-postgresql-service-snapshot), that restored service instance will by default include these extensions.
+
+If the snapshot contains extensions that GOV.UK PaaS no longer supports, the restored service instance will still include these extensions. However, you cannot create new service instances with unsupported extensions.
+
+#### Add or remove extensions when restoring a service instance
+
+When you restore a service instance, you can change the enabled extensions in that restored service instance. To do this, create your service with a new list of extensions:
```
bandesz: This should say you can enable additional extensions (not change them)

source/documentation/deploying_services/postgresql.md
```diff
@@ -271,6 +271,93 @@ cf update-service SERVICE_NAME -c '{"reboot": true, "force_failover": true}'
...
+cf create-service postgres PLAN NEW_SERVICE_NAME -c '{"restore_from_latest_snapshot_of": "GUID", "enabled_extensions": ["EXTENSION_1","EXTENSION_2",...,"EXTENSION_N"]}'
+```
+where:
+
+- `PLAN` is the plan used in the original service instance
+- `NEW_SERVICE_NAME` is a unique, descriptive name for the restored service instance
+- `GUID` is the global unique identifier of the backed up service instance
+- `EXTENSION_1...N` are the names of the extensions you want to enable
+
+You cannot use this method to enable an unsupported extension if that unsupported extension was not enabled in the original snapshot.
```
bandesz: I think this line is unnecessary as nothing indicates the possibility.

source/documentation/deploying_services/postgresql.md
AP-Hunt: The first paragraph is true as it stands. The second paragraph I agree is hypothetical, since we've not stopped supporting any extensions to date.

bandesz: (I referred only to the second paragraph)

source/documentation/deploying_services/postgresql.md
jonathanglassman: keeping in with `diable_extensions`

source/documentation/deploying_services/postgresql.md
jonathanglassman: removing

source/documentation/deploying_services/postgresql.md
jonathanglassman: Removed

jonathanglassman:
@AP-Hunt @bandesz please take a look at the changes

source/documentation/deploying_services/postgresql.md
```diff
@@ -271,6 +271,115 @@ cf update-service SERVICE_NAME -c '{"reboot": true, "force_failover": true}'
...
+
+<div style="height:1px;font-size:1px;">&nbsp;</div>
+
+|Extension|PostgreSQL version|
+|:---|:---|
+|pg_stat_statements|9.5, 10|
+
+<div style="height:1px;font-size:1px;">&nbsp;</div>
+
+You can enable or disable optional extensions for a PostgreSQL service instance by:
```
bandesz: The use of `by` here sounds strange, I think e.g. `during` or `when` would be more appropriate imo.

source/documentation/deploying_services/postgresql.md
```diff
@@ -271,6 +271,115 @@ cf update-service SERVICE_NAME -c '{"reboot": true, "force_failover": true}'
...
+where:
+
+- `SERVICE_NAME` is a unique descriptive name for this service instance
+- `EXTENSION_1...N` are the names of the additional optional extensions you want to enable
+
+#### Enable an optional extension in an existing service instance
+
+When you enable additional optional extensions in a service instance, you must also reboot that service instance.
+
+Run the following to enable additional optional extensions to a service instance:
```
bandesz: I suspect we should replace `to` with `in` or `on`.

source/documentation/deploying_services/postgresql.md
```diff
@@ -271,6 +271,115 @@ cf update-service SERVICE_NAME -c '{"reboot": true, "force_failover": true}'
...
+
+You can enable or disable optional extensions for a PostgreSQL service instance by:
+
+- creating a new service instance
+- updating an existing service instance
+- restoring a PostgreSQL service instance snapshot
+
+#### Create a new service instance
+
+You [create a new PostgreSQL service instance](#set-up-a-postgresql-service) by running `cf create-service`. You can enable additional optional extensions in this new service instance by running:
```
bandesz: `additional optional` - imo this is somewhat cluttering, I think just saying `additional` reads better

source/documentation/deploying_services/postgresql.md
```diff
@@ -271,6 +271,115 @@ cf update-service SERVICE_NAME -c '{"reboot": true, "force_failover": true}'
...
+```
+cf update-service SERVICE_NAME -c '{"enable_extensions": ["EXTENSION_1","EXTENSION_2",...,"EXTENSION_N"], "reboot": true}'
+```
+
+where:
+
+- `SERVICE_NAME` is a unique descriptive name for this service instance
+- `EXTENSION_1...N` are the names of the additional optional extensions you want to enable
+
+For example, your PostgreSQL service instance is named `my-pg-service` and has 3 mandatory extensions enabled, `uuid-ossp`, `postgis` and `citext`. Run the following to enable `pg_stat_statements`:
```
bandesz: I don't think we need to reiterate here again the mandatory extensions.

source/documentation/deploying_services/postgresql.md
```diff
@@ -271,6 +271,115 @@ cf update-service SERVICE_NAME -c '{"reboot": true, "force_failover": true}'
...
+
+|Extension|PostgreSQL version|
+|:---|:---|
+|uuid-ossp|9.5, 10|
+|postgis|9.5, 10|
+|citext|10|
+
+<div style="height:1px;font-size:1px;">&nbsp;</div>
+
+You cannot add or remove these mandatory extensions from a Postgresql service instance.
```
bandesz: IMO here it would be enough to say `You cannot remove these extensions from a Postgresql service instance`. (As you wouldn't add them).

source/documentation/deploying_services/postgresql.md
jonathanglassman: agree, using `when`

source/documentation/deploying_services/postgresql.md
jonathanglassman: will just use optional to link it to the first section's split between mandatory and optional

source/documentation/deploying_services/postgresql.md
jonathanglassman: I would like to keep mandatory in here in case the user does not read the first section which states the difference between mandatory and optional

source/documentation/deploying_services/postgresql.md
jonathanglassman: agree

source/documentation/deploying_services/postgresql.md
```diff
@@ -271,6 +271,115 @@ cf update-service SERVICE_NAME -c '{"reboot": true, "force_failover": true}'
...
+
+<div style="height:1px;font-size:1px;">&nbsp;</div>
+
+You can enable or disable optional extensions for a PostgreSQL service instance when:
+
+- creating a new service instance
+- updating an existing service instance
+- restoring a PostgreSQL service instance snapshot
+
+#### Create a new service instance
```
bandesz: One last thing, probably `Add extensions when creating a service instance` would be more in-line with the other headings.

PR #205, opened 2019-04-08
Update how to make the shell match the buildpack environment

rjw1:
buildpacks now provide a single command to make the ssh session match the environment of the running app.
as documented in https://docs.cloudfoundry.org/devguide/deploy-apps/ssh-apps.html#ssh-env
which was changed in https://github.com/cloudfoundry/docs-dev-guide/commit/ec036e7115ec24f2b9f58ff8abb67bee08a9a43b

What
----

Describe what you have changed and why.

How to review
-------------

1. …

Who can review
--------------


richardTowers:
Thanks Bob, this looks much nicer.

I've checked that this does indeed work in our deployment. I don't think there's any need for a second pair of tech writer eyes or comms involvement in this case, since it's just simplifying a command.

Merging.

PR #206, opened 2019-04-12
[SUPPORT] Explain `apps.internal` DNS behaviour

tnwhitwell:
What
----

Add a warning to explain that `apps.internal` DNS resolution will be case-sensitive: apps will not be routable if called with the wrong-cased dns name.

How to review
-------------

1. Check that the amendment makes sense, and conveys the correct meaning

Who can review
--------------

Not me. @jonathanglassman, could you please take a look?

### DIFF
--- a/source/documentation/deploying_apps/index.erb
+++ b/source/documentation/deploying_apps/index.erb
@@ -144,7 +144,7 @@ where:
 - `ENVIRONMENT_VARIABLE_NAME` is the [environment variable](#environment-variables) your public app reads (for example `PRIVATE_APP_URL`)
 - `PRIVATE_APPNAME` is the name of your private app
 
-<%= warning_text('When an app is accessed via the `*.apps.internal` DNS name (ie. by another app), the DNS resolution is case-sensitive. An app with the FQDN `test.apps.internal` can only be accessed as `test.apps.internal`: `TesT.apps.InternaL` will not be resolveable.') %>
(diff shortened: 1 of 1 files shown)


source/documentation/deploying_apps/index.erb
```diff
@@ -144,6 +144,8 @@ where:
 - `ENVIRONMENT_VARIABLE_NAME` is the [environment variable](#environment-variables) your public app reads (for example `PRIVATE_APP_URL`)
 - `PRIVATE_APPNAME` is the name of your private app
 
+<%= warning_text('When an app accesses a private app using the `*.apps.internal` DNS, this DNS must exactly match the route defined in the private app manifest. This includes capitalisation. <br><br> For example, a public app tries to access a private app that has the route `test.apps.internal`. If the public app uses `TesT.apps.InternaL`, this will cause an error.') %>
```
richardTowers: I wonder if calling this out with warning text is overly loud - I think it would be pretty rare for people to screw up the casing of their URL...

tnwhitwell: That's a fair point.. @jonathanglassman, i don't suppose you know off-hand if `info_text` is a thing, do you?

richardTowers: I'm almost persuaded that this is so much of an edge case that it's not worth documenting at all.

wHo UsEs UpPeRcAsE UrLs?

jonathanglassman: We can either have warning text or normal text for this. 
- Blockquotes using `>` are for direct quotations only, as per the GOV.UK style guide. 
- Bold text is for UI text only.

This did come about because of a support ticket. Given that it is just the one ticket, you can make a case for it being an edge case, and therefore not worth documenting. On the other hand, it is good to iterate docs based on user support tickets. I have no strong feelings either way.

tnwhitwell: I tHiNk YoU hAvE a PoInT.

Happy to file this in the circular cabinet.

PR #207, opened 2019-04-15
[#163694261] Workaround tech-docs-gem bug by using HTML headings

AP-Hunt:
What
----
There is a bug [1] in tech-docs-gem that causes identical headings on different
pages to be generated with incorrect identifiers. The behaviour is correct for
headings on the same page, but not for different pages as here. Jon Hallam, Jon
Glassman, and I agreed that a workaround using HTML headings was suitable here.

[1] https://github.com/alphagov/tech-docs-template/issues/180

How to review
-------------

1. Code review
2. Preview the content locally ([see README](https://github.com/alphagov/paas-tech-docs#preview)) and check that it renders as expected.
3. Check that all the table of contents headings under "Deploy a backing or routing service" work, and go to the right part of the page

Who can review
--------------
@jonathanglassman


jonathanglassman:
@AP-Hunt Demonstrated to community so ok to merge

PR #208, opened 2019-04-16
Update enabled RDS docs to match reality

tlwr:
What
----

Update enabled RDS docs to match reality

We have more extensions enabled since this:

https://github.com/alphagov/paas-cf/pull/1845

How to review
-------------

Read the commit above and review the table

Who can review
--------------

Not @tlwr


jonathanglassman:
@tlwr I assume the process for enabling or disabling these optional extensions has not changed?

tlwr:
@jonathanglassman nothing has changed other than the available extensions

PR #216, opened 2019-05-14
Cache Control Header in cdn-route service

jonathanglassman:
What
----

Added warning to cdn-route service content stating that you must set a Cache-Control response header if you want to ensure that you can change the response coming from your custom domain.

How to review
-------------

1. Check it fulfils story: https://www.pivotaltracker.com/story/show/165995762

Who can review
--------------

Anyone but Jon Glassman

### DIFF
--- a/source/documentation/deploying_services/use_a_custom_domain.erb
+++ b/source/documentation/deploying_services/use_a_custom_domain.erb
@@ -22,6 +22,9 @@ You must configure the cdn-route service to use a subdomain. If you configure th
 
 Once you create a CDN service instance, you cannot update or delete the instance until you have completed the setup process. If you make a mistake that breaks the configuration, email GOV.UK PaaS support at [gov-uk-paas-support@digital.cabinet-office.gov.uk](mailto:gov-uk-paas-support@digital.cabinet-office.gov.uk) to delete the service instance.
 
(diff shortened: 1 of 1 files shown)


source/documentation/deploying_services/use_a_custom_domain.erb
```diff
@@ -22,6 +22,9 @@ You must configure the cdn-route service to use a subdomain. If you configure th
 
 Once you create a CDN service instance, you cannot update or delete the instance until you have completed the setup process. If you make a mistake that breaks the configuration, email GOV.UK PaaS support at [gov-uk-paas-support@digital.cabinet-office.gov.uk](mailto:gov-uk-paas-support@digital.cabinet-office.gov.uk) to delete the service instance.
 
+<%= warning_text('The cdn-route service automatically caches responses from custom domains. <br><br> When you want to change the response from your custom domain, you must set a Cache-Control response header to specify that the cdn-route service should cache responses for a finite amount of time.') %>
```
richardTowers: I think there's a bit more that we need to capture here, but I'm not 100% sure I understand all of cloud front's behaviour. I'm going to mark this as "Request Changes" until I've got an idea of what we should actually be saying.

LeePorte: Is it worth linking to the AWS docs on cloudfront expiration? https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Expiration.html

PR #220, opened 2019-06-06
Correct advice on enabling SSO to say GDS users can self-serve

AP-Hunt:
What
----
George Lund at GDS spotted our tech docs advice and the advice in our emails were mismatched. This commit fixes that.

How to review
-------------
1. Read the words

Who can review
--------------
Anyone


AP-Hunt:
Being **bold** and just merging this.

PR #221, opened 2019-06-07
Remove the phase from the header

LeePorte:
What
----

Removing the phase from the header brings consistency with the header on the product pages.

How to review
-------------

Who can review
--------------

Anyone


jonathanglassman:
LGTM

PR #231, opened 2019-06-24
Amazon S3 bucket external access

jonathanglassman:
What
----

Added content on how to access the Amazon S3 bucket from outside of the GOV.UK PaaS.

How to review
-------------

1. Check it fulfils the https://www.pivotaltracker.com/story/show/166837820 story.

Who can review
--------------

Anyone except Jon Glassman or Tom Whitwell.

### DIFF
--- a/source/documentation/deploying_services/s3.md
+++ b/source/documentation/deploying_services/s3.md
@@ -72,7 +72,7 @@ applications:
 
 By default, you can only [connect to an S3 bucket from your GOV.UK PaaS-hosted app](#connect-to-an-s3-bucket-from-your-app).
 
-To [connect to an S3 bucket from outside the GOV.UK PaaS](#connect-to-an-s3-bucket-from-outside-of-the-gov-uk-paas), you must create a service key.   
+To [connect to an S3 bucket from outside the GOV.UK PaaS](#connect-to-an-s3-bucket-from-outside-of-the-gov-uk-paas), you must create a service key.  
 
 ### Connect to an S3 bucket from your app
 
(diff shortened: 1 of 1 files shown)


source/documentation/deploying_services/s3.md
```diff
@@ -106,6 +110,51 @@ An example `VCAP_SERVICES` environment variable:
...
+    {
+     "aws_access_key_id": "AWS_ACCESS_KEY",
+     "aws_region": "eu-west-2",
+     "aws_secret_access_key": "AWS_SECRET_ACCESS_KEY",
+     "bucket_name": "BUCKET_NAME_GUID",
+     "deploy_env": ""
+    }
+    ```
+
+    Use these credentials to connect to the S3 bucket from outside of the GOV.UK PaaS.
```
tnwhitwell: The only thing here is that I wonder if we should point out that the region specific **must** be used here, as we are specifically creating the buckets in the same region as the PaaS.

I'll need to do some digging to find out what that actually is though

source/documentation/deploying_services/s3.md
```diff
@@ -106,6 +110,51 @@ An example `VCAP_SERVICES` environment variable:
...
+
+To connect to the S3 bucket from outside of the GOV.UK PaaS, you must create a service key and credentials that allow external access to the bucket.
+
+1. Run the following to create a service key and credentials that allows external access:
+
+    ```
+    cf create-service-key SERVICE_NAME SERVICE-KEY -c '{"allow_external_access": true}'
+    ```
+    where:
+    - `SERVICE_NAME` is a unique descriptive name for the S3 bucket
```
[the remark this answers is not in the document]
jonathanglassman: agree

source/documentation/deploying_services/s3.md
tnwhitwell: not a valid concern: `aws s3 ls s3://BUCKET_NAME/` works A-OK

source/documentation/deploying_services/s3.md
```diff
@@ -106,6 +110,51 @@ An example `VCAP_SERVICES` environment variable:
...
+    ```
+
+    You will see output similar to the following:
+
+    ```
+    {
+     "aws_access_key_id": "AWS_ACCESS_KEY",
+     "aws_region": "eu-west-2",
+     "aws_secret_access_key": "AWS_SECRET_ACCESS_KEY",
+     "bucket_name": "BUCKET_NAME_GUID",
```
tnwhitwell:
```suggestion
     "bucket_name": "BUCKET_NAME",
```

richardTowers:
Do we need a second pair of tech writer eyes? Or shall I merge?

jonathanglassman:
> Do we need a second pair of tech writer eyes? Or shall I merge?

@richardTowers yes it needs 2i. I'll send on for 2i now.

PR #233, opened 2019-07-17
Redraft responsibility model

jonathanglassman:
What
----

- Redrafted responsibility model content to make it clear that tenants need to redeploy their apps when using standard buildpacks, and also explicitly state tenant responsibilities vs GOV.UK PaaS responsibilities.
- Make responsibility model its own section.
- Remove table as doesn't add anything and is not accessible.

I've not stated that this is security responsibilities, as we don't have non-security responsibilities in there yet. Will change later once we add this content.

How to review
-------------

1. Check that it fulfils https://www.pivotaltracker.com/story/show/164698889

Who can review
--------------

Anyone excluding @richardTowers and @jonathanglassman


source/documentation/responsibility_model/responsibility_model.md
```diff
@@ -0,0 +1,70 @@
...
+When you use the GOV.UK PaaS to host your app, both you and the GOV.UK PaaS have responsibilities. These responsibilities change depending on whether you deploy your app using a:
+
+- [standard](#standard-buildpack-responsibilities) buildpack
+- [custom](#custom-buildpack-responsibilities) buildpack
+- [Docker image](#docker-image-responsibilities)
+
+Regardless of which buildpack or Docker image you use, the GOV.UK PaaS team is always responsible for:
+
+- keeping the infrastructure your app runs on secure
+- providing secure app runtimes
```
richardTowers: I think "secure app runtimes" is a bit misleading, as it sounds like "language runtime" which we use differently.

I'd be inclined to just drop the second point and say "GOV.UK PaaS team is always responsible for keeping the infrastructure your app runs on secure".

PR #234, opened 2019-07-18
Single sign-on update

jonathanglassman:
What
----

The single sign-on function previously only worked for people in GDS with a GDS email address. It now works for everyone with a Google or Microsoft account. This updated content reflects this.

How to review
-------------

1. Check it completes https://www.pivotaltracker.com/story/show/167345982

Who can review
--------------

### DIFF
--- a/source/documentation/getting_started/get_started.md
+++ b/source/documentation/getting_started/get_started.md
@@ -56,39 +56,50 @@ The default method to sign in to Cloud Foundry is to use your [GOV.UK PaaS accou
 
 When you have signed in, run `cf` in the command line to see all available commands.
 
-### Use the single sign-on function
+### Use single sign-on 
 
-You can sign in to Cloud Foundry using the single sign-on function. Using single sign-on makes managing your security, onboarding and leaving processes simpler.
+You can sign in to Cloud Foundry using the single sign-on function. 
 
-Single sign-on is only available to GOV.UK PaaS users with either a [Google account](https://myaccount.google.com/intro) or a [Microsoft account](https://account.microsoft.com/account).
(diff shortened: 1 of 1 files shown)


source/documentation/getting_started/get_started.md
```diff
@@ -56,39 +56,50 @@ The default method to sign in to Cloud Foundry is to use your [GOV.UK PaaS accou
...
+### Use single sign-on 
 
-You can sign in to Cloud Foundry using the single sign-on function. Using single sign-on makes managing your security, onboarding and leaving processes simpler.
+You can sign in to Cloud Foundry using the single sign-on function. 
 
-Single sign-on is only available to GOV.UK PaaS users with either a [Google account](https://myaccount.google.com/intro) or a [Microsoft account](https://account.microsoft.com/account).
+Using single sign-on makes managing your security, joining and leaving processes simpler by reducing the number of passwords or accounts you have to manage.
 
-The email attached to your GOV.UK PaaS account must be the same as the email attached to your Google or Microsoft account.
+You must have either a [Google](https://myaccount.google.com/intro) or [Microsoft](https://account.microsoft.com/account) email address to use single sign-on. The email address must be the same as the one you use to sign into your GOV.UK PaaS account.
```
richardTowers: > The email address must be the same as the one you use to sign into your GOV.UK PaaS account.

Is this actually true? @tlwr?

tlwr: This used to be the case for Google

For Microsoft I do not think it is the case

I assume Google will no longer be the case after we enable it for everyone

richardTowers: That's my understanding too. A bit confusing to have people use accounts with different emails, but I think it will be possible.

Maybe it's better to have this sentence in the docs, even if it's not technically true?

PR #243, opened 2019-08-12
Correct PostgreSQL apps to PostgreSQL services

46bit:
What
----

Changes one word. PostgreSQL is a service, and the sentence is referring to what service plan it should use.

How to review
-------------

Read and consider.

Who can review
--------------

Not @46bit


jonathanglassman:
thanks for spotting and correcting @46bit

PR #245, opened 2019-08-13
Do not suggest updating CDN domains unless needed

tlwr:
What
----

The broker does not require all the configuration parameters anymore.

The broker needs a new DNS challenge whenever the value for domain has
changed, add a warning about this.

How to review
-------------

1. …

Who can review
--------------

Not @tlwr

### DIFF
--- a/source/documentation/deploying_services/use_a_custom_domain.erb
+++ b/source/documentation/deploying_services/use_a_custom_domain.erb
@@ -137,15 +137,9 @@ You can associate up to 100 subdomains with a single cdn-route service instance.
     ```
     cf update-service custom-domains-production \
       -c '{"domain": "www.example.com,www.example.net,www.example.org"}'
-    ```
+      ```
 
-    If you have previously customised the [cookies](#disabling-forwarding-cookies) or the [headers](#forwarding-headers) forwarded by your cdn-route service, you must include those as configuration parameters in the `cf update-service` command. For example:
-
-    ```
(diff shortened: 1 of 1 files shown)


source/documentation/deploying_services/use_a_custom_domain.erb
```diff
@@ -137,15 +137,9 @@ You can associate up to 100 subdomains with a single cdn-route service instance.
...
-
-    If you have previously customised the [cookies](#disabling-forwarding-cookies) or the [headers](#forwarding-headers) forwarded by your cdn-route service, you must include those as configuration parameters in the `cf update-service` command. For example:
+      ```
 
-    ```
-    cf update-service custom-domains-production \
-      -c '{"domain": "www.example.com,www.example.net,www.example.org",
-           "cookies": false, "headers": ["Accept", "Authorization"]}'
-    ```
+<%= warning_text('Whenever you update the "domain" configuration parameter you will have to update your DNS records, regardless of whether the value has changed.') %>
```
jonathanglassman: recommend changing "you will have to" to "you must" so it doesn't sound like you have to do this at some unspecified point in the future

jonathanglassman:
LGTM, just left one small comment

source/documentation/deploying_services/use_a_custom_domain.erb
tlwr: Force pushed correction, 🙇

PR #248, opened 2019-08-21
Add documentation for accessing public S3 buckets

gidsg:
What
----
Previously although it was described how to create public buckets, it was not described how to publicly access objects in those buckets.

How to review
-------------

1. Check the content makes sense

Who can review
--------------

Anyone but me

### DIFF
--- a/source/documentation/deploying_services/s3.md
+++ b/source/documentation/deploying_services/s3.md
@@ -44,6 +44,14 @@ cf create-service aws-s3-bucket default SERVICE_NAME -c '{"public_bucket":true}'
 ```
 where `SERVICE_NAME` is a unique descriptive name for this S3 bucket.
 
+#### Access objects in a public AWS S3 bucket
+
+Objects in a public S3 bucket will be available at:
+```
+https://BUCKET_NAME.s3.amazonaws.com/KEY
+```
+Where `BUCKET_NAME` is the `bucket_name` value available in [Environment Variables](/deploying_apps.html#system-provided-environment-variables) once you have [Bound an AWS S3 bucket to your app](/deploying_services/s3/#bind-an-aws-s3-bucket-to-your-app) and `KEY` is the key you gave to your object when performing `s3:PutObject`.
+
 ### Bind an AWS S3 bucket to your app
 
 You must bind the S3 bucket to your app so you can get credentials to use the [AWS S3 API](https://docs.aws.amazon.com/cli/latest/reference/s3api/) to read from and write to the bucket.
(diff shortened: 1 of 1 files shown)


tlwr:
Gideon thanks for raising this, this content looks good to me, but I've assigned it to Jon to take a look at

jonathanglassman:
@gidsg thanks for raising this. Content looks good although I do have some feedback. 

Given that you must first bind an AWS S3 bucket to your app to make this work, I don't think it should come before this section. Additionally, it seems to fit into using the service rather than setting up the service, especially given that it references `s3:PutObject` which is in the `Use the service` section.

I would therefore put the content into the `Use the service` section, incorporating a couple of changes to bring it in line with style:

"
<h2 id="use-the-service">Use the service</h2>

...

<div style="height:1px;font-size:1px;">&nbsp;</div>
|Description|Command|Permission|
|:---|:---|:---|
|Download an object from S3 bucket|`s3:GetObject`|`read-write`, `read-only`|
|Upload an object to S3 bucket|`s3:PutObject`|`read-write`|
|Delete an object from S3 bucket|`s3:DeleteObject`|`read-write`|
|Get AWS region of S3 bucket|`s3:GetBucketLocation`|`read-write`, `read-only`|
|List objects in S3 bucket|`s3:ListBucket`|`read-write`, `read-only`|
<div style="height:1px;font-size:1px;">&nbsp;</div>

### Access objects in a public AWS S3 bucket

Once you have uploaded an object to a [public S3 bucket](#provision-a-public-aws-s3-bucket) and [bound that bucket to your app](/deploying_services/s3/#bind-an-aws-s3-bucket-to-your-app), you can find that object at:

```
https://BUCKET_NAME.s3.amazonaws.com/KEY
```
where:

- `BUCKET_NAME` is the `bucket_name` value in the bucket [environment variables](/deploying_apps.html#system-provided-environment-variables) 
- `KEY` is the key you gave to the object when uploading that object to the bucket


"

Also, to clarify, is `KEY` the same as `SERVICE_KEY` in [this section](https://docs.cloud.service.gov.uk/deploying_services/s3/#connect-to-an-s3-bucket-from-outside-of-the-gov-uk-paas), or different?

gidsg:
@jonathanglassman `KEY` is different than `SERVICE_KEY`.

jonathanglassman:
@gidsg thanks. What about the rest of my proposed changes?

gidsg:
@jonathanglassman the suggested changes look fine to me, except if we are moving it to the _Use the service_ section, maybe we can remove the clause
> and bound that bucket to your app

as you would not be able to upload a file without doing this first so maybe that is a superfluous instruction now?

jonathanglassman:
> @jonathanglassman the suggested changes look fine to me, except if we are moving it to the _Use the service_ section, maybe we can remove the clause
> 
> > and bound that bucket to your app
> 
> as you would not be able to upload a file without doing this first so maybe that is a superfluous instruction now?

@gidsg fair enough, agree that we don't need "and bound that bucket to your app"

gidsg:
@jonathanglassman I have made the suggested changes. Please may you take another look?

jonathanglassman:
@gidsg LGTM

PR #268, opened 2019-10-23
Update buildpacks documentation

brenetic:
What
----

This adjusts the documentation buildpacks syntax to avoid getting deprecation warnings:

`Deprecation warning: Use of 'buildpack' attribute in manifest is deprecated in favor of 'buildpacks'`.

Who can review
--------------

Any dev or tech writer.

### DIFF
--- a/source/documentation/deploying_apps/buildpacks.md
+++ b/source/documentation/deploying_apps/buildpacks.md
@@ -52,7 +52,8 @@ You can set application attribute options in a manifest.yml file in the director
 ```
 ---
   ...
-  buildpack: buildpack_URL
+  buildpacks:
+    - buildpack_URL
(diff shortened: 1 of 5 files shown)


richardTowers:
In principal I would definitely agree with making this change, but beware! We're still [recommending](https://docs.cloud.service.gov.uk/managing_apps.html#404s-after-commands-that-restart-your-app) the blue-green-deployment plugin, which doesn't support the new non-deprecated syntax.

We should probably document `v3-zdt-push` as the recommended approach instead, then we could stop recommending deprecated manifest properties.

brenetic:
Ah ok fair enough. In which case I'll close this for now. It can be reopened if it's ever needed.

PR #280, opened 2019-11-27
Add backing service metrics content

jonathanglassman:
What
----

Added backing service metrics content.

How to review
-------------

1. Check it fulfils story https://www.pivotaltracker.com/story/show/169430669

Who can review
--------------

Anyone except Andy Hunt and Jon Glassman

### DIFF
--- a/source/documentation/monitoring_apps/metrics.md
+++ b/source/documentation/monitoring_apps/metrics.md
@@ -126,13 +126,13 @@ For more information about monitoring apps, see [Monitoring the status of your s
 
 ## Backing service metrics
 
-If you use the [PostgreSQL](deploying_services/postgresql/#postgresql), [MySQL](deploying_services/mysql/#mysql) or [Redis](deploying_services/redis/#redis) backing service, you can view metrics on those backing services in the GOV.UK PaaS admin tool.
+If you use the [PostgreSQL](deploying_services/postgresql/#postgresql), [MySQL](deploying_services/mysql/#mysql) or [Redis](deploying_services/redis/#redis) backing services, you can view metrics for those backing services in the GOV.UK PaaS admin tool.
 
-When you create your backing service instance, you can access its metrics with no additional work. Your backing service plan does not affect which metrics you can view.
+When you create your backing service instance, you can access the service instance’s metrics with no extra work. Your backing service plan does not affect which metrics you can view. You can view metrics for any date range within the last 60 days.
 
-You can view metrics up to 60 calendar days before today's date. You can select any date range within this time period.
+Backing services instances exist in a [space](orgs_spaces_users.html#spaces) within an [org](orgs_spaces_users.html#organisations). You must be able to view a space in the GOV.UK PaaS admin tool to view metrics for backing service instances in that space.
 
-Backing services are attached to a [space](orgs_spaces_users.html#spaces) within an [org](orgs_spaces_users.html#organisations). You must be able to view a space in the GOV.UK PaaS admin tool to view metrics for backing services in that space.
+### View backing service metrics
 
 1. Sign into the GOV.UK PaaS admin tool for [London](https://admin.london.cloud.service.gov.uk/) or [Ireland](https://admin.cloud.service.gov.uk/).
 1. Select the __Organisation__ and __Space__.
@@ -141,11 +141,9 @@ Backing services are attached to a [space](orgs_spaces_users.html#spaces) within
 
 ### Metrics definitions - PostgreSQL and MySQL
 
-PostgreSQL and MySQL are relational database management systems.
-
 #### Free disk space
 
-How much hard disk space your database has remaining. Your database will stop working if it runs out of disk space. Upgrade your service plan to get more disk space.
+How much hard disk space your database has left. Your database will stop working if it runs out of disk space. Upgrade your service plan to get more disk space.
 
 #### CPU utilisation
 
@@ -153,21 +151,21 @@ How much computational work your database is doing. If you think your CPU utilis
 
 #### Open connections
 
-How many open connections there are to your database. Unexpectedly high values may indicate problems with your apps' connection management, or that you need to upgrade your service plan. Unexpectedly low values may indicate the database is unavailable, or that your apps cannot connect to the database.
+How many open connections there are to your database. If the values are unexpectedly high, you may need to optimise how your app connects to the database, or upgrade your service plan. Unexpectedly low values may indicate the database is unavailable, or your apps cannot connect to the database.
 
-#### Available Memory
+#### Available memory
 
-How much memory the virtual machine your database is running on has remaining. Persistently low values may indicate you need to optimise your database queries or upgrade your service plan.
+How much memory the virtual machine your database is running on has left. If values are persistently low, you may need to optimise your database queries, or upgrade your service plan.
 
 #### Read and write operations per second
 
-How many read and write operations per second (IOPS) your database is performing per second. Databases have a limit of 3 IOPS per gigabyte of database hard disk space. This limit applies to the sum of the read and write operations. For example, a 100 gigabyte database has a 300 IOPS limit. If your database is close to its IOPS limit, you can upgrade your service plan.
+How many read and write operations per second (IOPS) your database is performing per second. Databases have a limit of 3 IOPS per gigabyte of database hard disk space. This limit applies to the sum of the read and write operations. 
+
+For example, a 100 gigabyte database has a 300 IOPS limit. If your database is close to its IOPS limit, you can upgrade your service plan.
 
 See the [documentation on Amazon RDS database instance storage](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Storage.html#CHAP_Storage.IO.Credits) for more information.
 
 ### Metrics definitions - Redis
-
-Redis is an open source in-memory data store.
 
 #### CPU utilisation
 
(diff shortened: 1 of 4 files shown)


source/documentation/monitoring_apps/metrics.md
```diff
@@ -123,3 +123,84 @@ If you are not receiving any metrics, check the [logs](/monitoring_apps.html#log
 ### More about monitoring
 
 For more information about monitoring apps, see [Monitoring the status of your service](https://www.gov.uk/service-manual/technology/monitoring-the-status-of-your-service) on the Service Manual.
+
+## Backing service metrics
+
+If you use the [PostgreSQL](deploying_services/postgresql/#postgresql), [MySQL](deploying_services/mysql/#mysql) or [Redis](deploying_services/redis/#redis) backing service, you can view metrics on those backing services in the GOV.UK PaaS admin tool.
```
AP-Hunt: > If you use ~the~ a [PostgreSQL](deploying_services/postgresql/#postgresql), [MySQL](deploying_services/mysql/#mysql) or [Redis](deploying_services/redis/#redis) backing service, you can view metrics on ~those backing services~ it in the GOV.UK PaaS admin tool.

Thoughts?

source/documentation/monitoring_apps/metrics.md
```diff
@@ -123,3 +123,84 @@ If you are not receiving any metrics, check the [logs](/monitoring_apps.html#log
...
+
+When you create your backing service instance and bind an app to that backing service, you can access that backing service's metrics for that app. Your backing service plan does not affect which metrics you can view.
```
AP-Hunt: Backing service metrics aren't app scoped. This should probably read like below, or something to that effect maybe?

> When you create your backing service instance, you can access its metrics with no additional work. Your backing service plan does not affect which metrics you can view.

source/documentation/monitoring_apps/metrics.md
```diff
@@ -123,3 +123,84 @@ If you are not receiving any metrics, check the [logs](/monitoring_apps.html#log
...
+
+How much computational work your database is doing. If you think your CPU utilisation is too high, you can optimise your database queries or upgrade your service plan.
+
+#### Open connections
+
+How many open connections there are to your database. Unexpectedly high values may indicate problems with your apps' connection management, or that you need to upgrade your service plan. Unexpectedly low values may indicate the database is unavailable, or that your apps cannot connect to the database.
+
+#### Available Memory
+
+How much memory the virtual machine your database is running on has remaining. Unexpectedly low values may indicate you need to optimise your database queries or upgrade your service plan.
```
AP-Hunt: I'm not 100% sure how query optimisation would affect memory usage. It's maybe better to say "Persistently low values" to communicate that it's OK for it to dip occasionally, but that a long running low figure is maybe bad.

source/documentation/monitoring_apps/metrics.md
```diff
@@ -123,3 +123,84 @@ If you are not receiving any metrics, check the [logs](/monitoring_apps.html#log
...
+
+#### Read and write operations per second
+
+How many read and write operations per second (IOPS) your database is performing per second. Databases have a limit of 3 IOPS per gigabyte of database hard disk size. For example, a 100 gigabyte database has a 300 IOPS limit. If your database is close to its IOPS limit, you can upgrade your service plan.
```
AP-Hunt: >per gigabyte of hard disk size

We say disk space elsewhere, so I think we should be consistent

AP-Hunt: > If your database is close to its IOPS limit

To be clearer for tenants, we should make it explicit that the IOPS limit is the sum of read and write.

source/documentation/monitoring_apps/metrics.md
```diff
@@ -123,3 +123,84 @@ If you are not receiving any metrics, check the [logs](/monitoring_apps.html#log
...
+
+See the [documentation on Amazon RDS database instance storage](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Storage.html#CHAP_Storage.IO.Credits) for more information.
+
+### Metrics definitions - Redis
+
+Redis is an open source in-memory data store.
+
+#### CPU utilisation
+
+How much computational work your Redis service instance is doing. If you think your CPU utilisation is too high, you can optimise your database queries or upgrade your service plan.
```
AP-Hunt: Redis doesn't have queries in the same way as a relational database. A clearer formulation might be "optimise how you read data from, and write data to, Redis"

source/documentation/monitoring_apps/metrics.md
```diff
@@ -123,3 +123,84 @@ If you are not receiving any metrics, check the [logs](/monitoring_apps.html#log
...
+
+#### Memory used
+
+How much memory your Redis service instance is using to run itself and to store your app data. Memory limit is pre-set based on service plan and you cannot change this memory limit. If your service instance reaches its memory limit, it will start [evicting data keys](monihttps://redis.io/topics/lru-cache). This may indicate that you need to review the size of the data you're storing, or that you need to upgrade your service plan.
```
AP-Hunt: > (monihttps://redis.io/topics/lru-cache).

Typo

source/documentation/monitoring_apps/metrics.md
```diff
@@ -123,3 +123,84 @@ If you are not receiving any metrics, check the [logs](/monitoring_apps.html#log
...
+
+#### Swap memory used
+
+If your Redis service instance is running low on memory, it will start to swap memory onto the hard disk. To reduce memory swapping, you can reduce the number of apps using your service instance or upgrade your service plan.
```
AP-Hunt: > To reduce memory swapping, you can reduce the number of apps using your service

It's more about making an effort to reduce the amount of memory being used than reducing the number of apps using the service.

source/documentation/monitoring_apps/metrics.md
```diff
@@ -123,3 +123,84 @@ If you are not receiving any metrics, check the [logs](/monitoring_apps.html#log
...
+
+#### Key evictions
+
+Your Redis service instance will delete or [evict keys](https://redis.io/topics/lru-cache) when it reaches its memory limit. Memory limit is pre-set based on service plan and you cannot change this memory limit.
+
+Redis instances on the GOV.UK PaaS use the [volatile-lru policy](https://redis.io/topics/lru-cache#eviction-policies). This means Redis instances can only evict keys that have an expire set or expiry date. Your Redis service instance will try to evict less recently used keys first. If your service instance cannot evict any keys, it will return errors when executing commands that increase memory use. Upgrade your service plan to reduce key eviction.
```
AP-Hunt: > keys that have an expire set or expiry date

probably better read as

> keys that have an expiry time set

source/documentation/monitoring_apps/metrics.md
jonathanglassman: Our style is that we try to minimise the use of "it" to avoid ambiguity, especially when there are several things "it" could refer to. Means the text is a bit clunky sometimes, but it is explicitly clear.

source/documentation/monitoring_apps/metrics.md
jonathanglassman: agree

source/documentation/monitoring_apps/metrics.md
jonathanglassman: agree

source/documentation/monitoring_apps/metrics.md
jonathanglassman: agree x2

source/documentation/monitoring_apps/metrics.md
jonathanglassman: agree

source/documentation/monitoring_apps/metrics.md
jonathanglassman: whoops!

source/documentation/monitoring_apps/metrics.md
jonathanglassman: will change

source/documentation/monitoring_apps/metrics.md
jonathanglassman: agree

source/documentation/monitoring_apps/metrics.md
```diff
@@ -171,21 +171,21 @@ Redis is an open source in-memory data store.
...
+How much memory your Redis service instance is using to run itself and to store your app data. Memory limit is pre-set based on service plan and you cannot change this memory limit. If your service instance reaches its memory limit, it will start [evicting data keys](https://redis.io/topics/lru-cache). This may indicate that you need to review the size of the data you're storing, or that you need to upgrade your service plan.
 
 #### Swap memory used
 
-If your Redis service instance is running low on memory, it will start to swap memory onto the hard disk. To reduce memory swapping, you can reduce the number of apps using your service instance or upgrade your service plan.
+If your Redis service instance is running low on memory, it will start to swap memory onto the hard disk. To reduce memory swapping, you can reduce the amount of memory your apps use or upgrade your service plan.
```
AP-Hunt: Apologies, looking at my previous comment I can see that it was ambiguous. I meant the amount of data being used by Redis.

jonathanglassman: do we need to tell them how to do this?

AP-Hunt: I don't think so. We explained that in the bit about memory consumption iirc

jonathanglassman: ok

source/documentation/monitoring_apps/metrics.md
```diff
@@ -123,3 +123,84 @@ If you are not receiving any metrics, check the [logs](/monitoring_apps.html#log
...
 
 For more information about monitoring apps, see [Monitoring the status of your service](https://www.gov.uk/service-manual/technology/monitoring-the-status-of-your-service) on the Service Manual.
+
+## Backing service metrics
+
+If you use the [PostgreSQL](deploying_services/postgresql/#postgresql), [MySQL](deploying_services/mysql/#mysql) or [Redis](deploying_services/redis/#redis) backing service, you can view metrics on those backing services in the GOV.UK PaaS admin tool.
+
+When you create your backing service instance, you can access its metrics with no additional work. Your backing service plan does not affect which metrics you can view.
+
+You can view metrics up to 60 calendar days before today's date. You can select any date range within this time period.
```
richardTowers: I think we've enabled more than 60 days for now - it's ~1 year at the moment. Perhaps this is here as a bit of future proofing though?

AP-Hunt: Jon and I did some digging around in CloudWatch and found that metrics seemed only to go as far back as 60 days. Although we only looked at really long lived DBs like bosh.

jonathanglassman: As per discussion with @richardTowers will change to 1 year

PR #295, opened 2020-03-04
Document the Kibana alongside every Elasticsearch

46bit:
What
----

A tenant asked us about getting Kibana access to their Elasticsearch. We discovered Aiven already provide one. This commit saves the instructions that we gave that tenant so that others can benefit too.

Here's what this content looks like:

> Screenshot 2020-03-04 at 14 34 30

How to review
-------------

Read through the instructions and see if they make sense.

Who can review
--------------

Not @46bit


AP-Hunt:
Since this is changing user behaviour, does it need to go through 2i @46bit ?

46bit:
I'm going to :shipit: ship this because it is a relatively minor change and I would rather close it than spend awhile doing 2i. Maybe that's bad of me.

PR #299, opened 2020-05-04
Docker images are cached by tag, and reusing a tag can cause errors.

AP-Hunt:
What
----
Docker images are cached by tag, and reusing a tag can cause errors.

Advises tenants not to do this, and instead to use a unique tag. Suggests what
that tag could be.

How to review
-------------

1. Check you agree with the gist of what's being said

Who can review
--------------
Anyone


AP-Hunt:
> I don't suppose we could resolve to using Digests?

Possibly, but I don't know how. Worth a look in to when I have some free time, probably.

PR #301, opened 2020-05-20
Update non-PaaS postgres transfer guidance

tlwr:
What
----

@saliceti reported in x-gov slack that these flags were needed to
restore a paas database remotely

i have verified that without these flags, users see role errors (which
may be fine to ignore, but better to suppress them to make them less
confusing)

How to review
-------------

[Check flags exist](https://www.postgresql.org/docs/9.5/app-pgdump.html)

Who can review
--------------

Not @tlwr


saliceti:
Is it also worth mentioning expected errors about reassign_owned, extensions and spatial_ref_sys ?

tlwr:
Superseded

PR #343, opened 2020-09-14
[#171281178] Document metric exporting to prometheus

46bit:
What
----

Does three main things:

1. Separates the monitoring documentation into different pages for apps and services
2. Adds instructions on how to get backing service metrics into Prometheus
3. Removes metric descriptions that now exist inline with the metric graphs in `paas-admin`

See the commit messages for more details.

How to review
-------------

1. [ ] Preview the content locally ([see README](https://github.com/alphagov/paas-tech-docs#preview)) and check that it renders as expected.
1. [x] Manually check that any removed or renamed anchor tags are not in use ([linkchecker](https://github.com/linkcheck/linkchecker) doesn't do this).

Who can review
--------------

Not @46bit


46bit:
Now that I've removed the metrics definitions, the "Monitoring backing services" page is quite short. I suggest we keep it that way, as the "Monitoring apps" section is huge.

I think we need to be able to have separate pages where necessary--for instance, instructions for setting up Logit. These long sets of instructions pollute the pages and make it impossible to find anything by scrolling.

46bit:
The Monitoring Apps docs could use some love, but I'm not going there right now. I've filed #344 for one of the most obvious issues with them.

paroxp (approved):

PR #357, opened 2020-11-16
remove 'yourself'

gidsg:
The 'yourself' here doesn't make sense as both restore methods are now self-service.

What
----

I found the documentation confusing as it implied one method was more self-service than the other.

How to review
-------------

Who can review
--------------

not me


kr8n3r (approved):

kr8n3r:
Thanks Gideon

PR #362, opened 2020-12-17
Add information about upgrading from postgres 9.5

LeePorte:
What
----

In order to make upgrading a smoother process it makes sense to document
this in our tech docs rather than asking people to check for old emails.
It also makes sense to pull this up to an appropriate level so that it
is surfaces on the side menu.

How to review
-------------

1. Check the words added

Who can review
--------------

Probably @markpeterbuckley


markpeterbuckley:
Can this go at the end of https://docs.cloud.service.gov.uk/deploying_services/postgresql/#upgrade-postgresql-service-plan ? Think this is in the wrong place.

PR #369, opened 2021-01-26
[#176574936] `postgis` is no longer enabled by default for any version of Postgres

AP-Hunt:
What
----

See https://github.com/alphagov/paas-cf/pull/2583 

How to review
-------------
Check that there's nowhere else in the docs that suggests `postgis` is enabled by default

Who can review
--------------
Claire McNally, and one other


cmcnallygds (approved):
Looks good

PR #384, opened 2021-05-17
Add StandardJS linting workflow

0atman:
What
----

Our style of choice is StandardJS. Though there's not much JS here, it should be Standard.

### pre-merge

- [x] Add standard package to npm
- [x] Add `npm lint` and `lint:fix` targets
- [x] Auto-run in a GH action
- [x] Auto-fix the js in this repo, using correct eslint config for our usecase (which for us is `eslint-disable no-var, prefer-const`)

### post merge

- [ ] Check tech docs JS works once built

How to review
-------------

1. The tech docs should still build and look the same.
2. `npm install && npm test` should succeed.

Who can review
--------------

Any devs

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,22 @@
+{
+  "dependencies": {
+    "standard": "^16.0.3"
+  },
+  "scripts": {
+    "test": "npm run lint",
+    "lint": "standard",
+    "lint:fix": "standard --fix"
+  },
+  "standard": {
+    "ignore": [
+      "out/*"
+    ],
+    "env": [
+      "es6",
+      "browser"
+    ],
+    "globals": [
+      "dataLayer"
+    ]
+  }
+}
--- a/package.json
+++ b/package.json
@@ -8,11 +8,7 @@
     "lint:fix": "standard --fix"
   },
   "standard": {
-    "ignore": [
-      "out/*"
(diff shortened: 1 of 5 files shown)


0atman:
Looks like the standard tests work :ok_hand: 
Screenshot from 2021-05-17 15-14-26

paroxp (approved):

package.json
```diff
@@ -0,0 +1,22 @@
...
+    "standard": "^16.0.3"
+  },
+  "scripts": {
+    "test": "npm run lint",
+    "lint": "standard",
+    "lint:fix": "standard --fix"
+  },
+  "standard": {
+    "ignore": [
+      "out/*"
```
kr8n3r, line 12: `build` -  (`out` is from product pages)

package.json
```diff
@@ -0,0 +1,22 @@
...
+    ],
+    "env": [
+      "es6",
```
kr8n3r, line 15: JS here will be all ES5 me thinks

kr8n3r (approved):
minor changes, otherwise looks good

package.json
0atman: hmm, no es5 option for standard, I'll remove it.

PR #394, opened 2021-07-14
Use full service name in the `title`

kr8n3r:
What
----

Tech docs template allows for a [`full_service_name`](https://tdt-documentation.london.cloudapps.digital/configure_project/global_configuration/#full-service-name)
that gets used in the title. We keep the `service_name` for use in the header, as header comes with the GOV.UK logo already

Before

Screenshot 2021-07-14 at 09 35 17

After

Screenshot 2021-07-14 at 09 31 59

How to review
-------------

Who can review
--------------
not @kr8n3r

### DIFF
--- a/config/tech-docs.yml
+++ b/config/tech-docs.yml
@@ -4,6 +4,7 @@ host: https://docs.cloud.service.gov.uk
 # Header-related options
 show_govuk_logo: true
 service_name: Platform as a Service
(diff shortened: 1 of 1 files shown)


0atman (approved):
Works for me locally :ok_hand: 
Screenshot from 2021-07-14 10-10-12

PR #408, opened 2021-09-17
Update postgresql section (including some mysql levelling up)

risicle:
What
----

Renamed `postgresql-hll` to `hll` as that's what it's actually available as, added postgresql 13 extension availability information, add basic note on high-iops plans for both postgresql and mysql.

How to review
-------------

:point_up: yeah that

Further questions
---------

Should I change all the examples to talk about postgres 13? (I'd prefer to wait until we know it's really solid for people TBH)

Should we add a paragraph talking about to what degree we (don't) support the extensions as we discussed previously?

Should we talk about some extensions having dependencies that need adding in the right order?


tomrosier (approved):

source/documentation/deploying_services/mysql.md
```diff
@@ -345,6 +345,10 @@ Refer to the [Amazon RDS documentation on read replicas](https://aws.amazon.com/
 
 GOV.UK PaaS does not currently support read replicas, but if you think you would find them useful, please contact us at [gov-uk-paas-support@digital.cabinet-office.gov.uk](mailto:gov-uk-paas-support@digital.cabinet-office.gov.uk), providing details of your use case.
 
+#### High IOPS plans - MySQL
+
+For services which require heavy disk throughput, GOV.UK PaaS has `high-iops` variants of its MySQL service plans available.
```
cmcnallygds, line 350: Will it be clear to the user what we mean by 'heavy disk throughput'?

risicle: I don't know, but I didn't want to use the term that AWS use to describe "provisioned IO" storage, which is

> For a production application that requires fast and consistent I/O performance

because that might imply that we use "provisioned IO" for this feature, and we don't, we use a hack with gp2 storage instead, (which doesn't necessarily give the same consistency guarantees).

Anyone got any better suggestions?

0atman: I think that'll very comprehensible even to novice users :ok_hand:

AP-Hunt: What about "services which have a large volume of database reads and writes"?

source/documentation/deploying_services/postgresql.md.erb
```diff
@@ -345,65 +345,71 @@ GOV.UK PaaS supports the following optional extensions, but they are not enabled
 
 |Extension|PostgreSQL version|
```
AP-Hunt, line 346: I'm starting to wonder if we couldn't make this table better

Extension | 10 | 11 | 12 | 13 
-- | -- | -- | -- | -- | -- | -- | --
address_standardizer|✓|✓|✓|✓
chkpass|✓|✓|✓|𐄂

(I can't make Github's markdown parser happy with that table so sod it, syntax only)

risicle:
Anyone got any thoughts on the *Further questions* I posted up top :point_up: ? These are the things I'd really like opinions on.

source/documentation/deploying_services/postgresql.md.erb
risicle: Ug fighting with table syntaxes. A little digging suggests our markdown parser is `kramdown`? https://kramdown.gettalong.org/syntax.html#tables

PR #420, opened 2021-11-02
Correct flag used in CF marketplace command

AP-Hunt:
What
----

The `-s` flag was correct in CF CLI v6, but changed to `-e` in
CF CLI v7.


schmie (approved):
and I thought I was crazy :sweat_smile:

PR #423, opened 2021-11-29
Update docs to include OpenSearch

cadmiumcat:
What
----

We are making OpenSearch available to tenants, but we need some docs so that they know how to use it 🤓 

My assumptions:

- backups work similarly to how they do for elasticsearch
- we export OpenSearch metrics on request (as we do for elasticsearch)
- I don't need to add a [compatibility note for elasticsearch-py](https://github.com/alphagov/paas-tech-docs/blob/07648983d408de2029d9db5a9996b14df9b124a2/source/documentation/deploying_services/elasticsearch.md#compatibility-notes)

How to review
-------------

Describe the steps required to test the changes.

1. Preview the content locally ([see README](https://github.com/alphagov/paas-tech-docs#preview)) and check that it renders as expected.
2. Read the docs,  make sure they make sense
3. Check I haven't missed adding OpenSearch references somewhere

Who can review
--------------

Anyone

PR #424, opened 2021-12-02
Updates docs to reflect changes from OpenSearch docs

cadmiumcat:
What
----

The PR for OpenSearch docs #423 had some suggestions that also applied to the ES docs. I'm putting them in here so that we don't forget them 

How to review
-------------

2. Read the docs,  make sure they make sense
3. Check the changes correspond to the suggestions in #423 

Who can review
--------------

Anyone

### DIFF
--- a/source/documentation/deploying_services/elasticsearch.md
+++ b/source/documentation/deploying_services/elasticsearch.md
@@ -140,7 +140,7 @@ where `SERVICE_NAME` is a unique descriptive name for this service instance, and
 cf update-service my-elasticsearch-service -p small-ha-7.x
 ```
 
-The plan upgrade will start immediately and finish within an hour. You can check the status of the upgrade by running `cf services`.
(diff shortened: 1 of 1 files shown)


source/documentation/deploying_services/elasticsearch.md
```diff
@@ -140,7 +140,7 @@ where `SERVICE_NAME` is a unique descriptive name for this service instance, and
 cf update-service my-elasticsearch-service -p small-ha-7.x
 ```
 
-The plan upgrade will start immediately and finish within an hour. You can check the status of the upgrade by running `cf services`.
+The plan upgrade will start immediately and finish within an hour, except for large plans which may take over an hour. You can check the status of the upgrade by running `cf services`.
```
cmcnallygds, line 143: It might be worth giving an example of what we mean by 'large' - what kind of size, exactly?

AP-Hunt: I think this refers to `large` and `xlarge` plans.

AP-Hunt:
Folding this in to #435

PR #427, opened 2021-12-09
[#180331598] Flesh out our AWS Presigned URLs documentation

AP-Hunt:
What
----

We deliberately don't give any usage examples here, because a useful example
would be using an AWS SDK for the user's programming language, and we don't
know what that is.

Instead, we suggest what they're good for and direct people to look at the relevant SDK.

How to review
-------------

1. Do you agree with the approach?

### DIFF
--- a/source/documentation/deploying_services/s3.md
+++ b/source/documentation/deploying_services/s3.md
@@ -74,7 +74,7 @@ applications:
 
 By default, you can only [connect to an S3 bucket from your GOV.UK PaaS-hosted app](#connect-to-an-s3-bucket-from-your-app).
 
-To [connect to an S3 bucket from outside the GOV.UK PaaS](#connect-to-an-s3-bucket-from-outside-of-the-gov-uk-paas), you must create a service key.  
+To [connect to an S3 bucket from outside the GOV.UK PaaS](#connect-to-an-s3-bucket-from-outside-of-the-gov-uk-paas), you must create a service key.
 
 ### Connect to an S3 bucket from your app
 
@@ -110,7 +110,7 @@ An example `VCAP_SERVICES` environment variable:
  }
 }
 ```
-You must pass credentials from your `VCAP_SERVICES` environment variable to your AWS software development kit (SDK). Use the appropriate [Amazon SDK documentation](https://aws.amazon.com/developer/tools/#sdks) to do this.  
+You must pass credentials from your `VCAP_SERVICES` environment variable to your AWS software development kit (SDK). Use the appropriate [Amazon SDK documentation](https://aws.amazon.com/developer/tools/#sdks) to do this.
 
 ### Connect to an S3 bucket from outside of the GOV.UK PaaS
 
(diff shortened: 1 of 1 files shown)


source/documentation/deploying_services/s3.md
```diff
@@ -179,21 +179,25 @@ You can take different actions on an S3 bucket using the [AWS S3 API](https://do
...
+- `BUCKET_NAME` is the `bucket_name` value in the bucket [environment variables](/deploying_apps.html#system-provided-environment-variables)
 - `KEY` is the key you gave to the object when uploading that object to the bucket
 
 ### Access objects in a private AWS S3 bucket using presigned URLs
 
-You can [generate a presigned URL](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-presigned-url.html)
-that gives temporary access to an object in a private bucket.
+You can [generate a presigned URL](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-presigned-url.html) that gives temporary access to an object in a private bucket. Presigned URLs are useful in scenarios where you want to:
+
+* [grant limited access to a resource in the bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ShareObjectPreSignedURL.html) after validating the user's authorization (for example, downloading a signed document)
```
cadmiumcat, line 197: > after validating the user's authorization
Is this the user who creates the URL or the user who has temporary access? I assume the former, but I think it's not clear

Also:
```suggestion
* [grant limited access to a resource in the bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ShareObjectPreSignedURL.html) after validating the user's authorisation (for example, downloading a signed document)
```

AP-Hunt: > Is this the user who creates the URL or the user who has temporary access? I assume the former, but I think it's not clear

I meant the user of the application, and the application applying its domain logic to see if that user should be allowed to access the thing.

cadmiumcat (approved):
👍  this makes sense

source/documentation/deploying_services/s3.md
```diff
@@ -179,21 +179,25 @@ You can take different actions on an S3 bucket using the [AWS S3 API](https://do
...
 
-You can [generate a presigned URL](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-presigned-url.html)
-that gives temporary access to an object in a private bucket.
+You can [generate a presigned URL](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-presigned-url.html) that gives temporary access to an object in a private bucket. Presigned URLs are useful in scenarios where you want to:
+
+* [grant limited access to a resource in the bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ShareObjectPreSignedURL.html) after validating the user's authorisation (for example, downloading a signed document)
+* [allow users to upload files directly to the bucket](https://docs.aws.amazon.com/AmazonS3/latest/userguide/PresignedUrlUploadObject.html), without passing the file through your application (for example, large file uploads)
+
+You can generate a presigned URL in your application using the AWS SDK package for your application's programming language. You can also generate them with the AWS CLI tool by using [the S3 bucket credentials from a service key](/deploying_services/s3/#connect-to-an-s3-bucket-from-outside-of-the-gov-uk-paas).
```
cmcnallygds, line 200:
```suggestion
You can generate a presigned URL in your application using the AWS SDK package for your application's programming language. You can also generate them with the AWS CLI tool by using the [S3 bucket credentials from a service key](/deploying_services/s3/#connect-to-an-s3-bucket-from-outside-of-the-gov-uk-paas).
```

cmcnallygds (changes requested):
Have suggested a tiny change for link text

cmcnallygds (approved):

PR #428, opened 2022-01-05
opensearch restore: `restore_from_latest_backup_of` expects a guid

risicle:
https://www.pivotaltracker.com/story/show/180810438

What
----

Turns out you don't use the service name for restoring OS instances, you use the GUID. Really, we should probably flesh this out to be more like (a clone of?) the instructions we use for postgres, where we include details of how to discover a service's guid - but right now I just want to get these words switched to prevent this from being wrong.

How to review
-------------

1. …

Who can review
--------------

Not @risicle


AP-Hunt (approved):
Yes it expects the service GUID, not the service name: https://github.com/alphagov/paas-aiven-broker/blob/bf7bf8f13087970fc83f6843a259fa31367ee9c5/provider/provider.go#L402

PR #429, opened 2022-01-13
Explains how to check TLS version and ciphers

cadmiumcat:
What
----
Explains how to find out about TLS version and ciphers

How to review
-------------

1. Read and check that the new info makes sense

Who can review
--------------

Anyone

### DIFF
--- a/source/documentation/deploying_services/use_a_custom_domain.erb
+++ b/source/documentation/deploying_services/use_a_custom_domain.erb
@@ -450,6 +450,10 @@ The above example configures the CDN route to have a default time-to-live of 15 
 
 The CDN route service forwards cookie headers your app by default, so cookie-based authentication will work as expected. Other headers, such as HTTP auth, are stripped by default. If you need a different configuration, see the guidance on [forwarding headers](/deploying_services/use_a_custom_domain/#forwarding-headers).
 
+### Verifying protocols and cyphers
+
+You can find the TLS version and cyphers used on GOV.UK PaaS by using an SSL testing tool such as [SSL Labs](https://www.ssllabs.com/ssltest/). For your apps, use your custom domain as the hostname.
+
 ### Further information
 
 See the GOV.UK guidance on:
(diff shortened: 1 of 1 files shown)


source/documentation/deploying_services/use_a_custom_domain.erb
```diff
@@ -450,6 +450,10 @@ The above example configures the CDN route to have a default time-to-live of 15
 
 The CDN route service forwards cookie headers your app by default, so cookie-based authentication will work as expected. Other headers, such as HTTP auth, are stripped by default. If you need a different configuration, see the guidance on [forwarding headers](/deploying_services/use_a_custom_domain/#forwarding-headers).
 
+### Verifying protocols and cyphers
+
+You can find the TLS version and cyphers used on GOV.UK PaaS by using an SSL testing tool such as [SSL Labs](https://www.ssllabs.com/ssltest/). For your apps, check the hostname `cloudapps.digital` (or `london.cloudapps.digital` in the London region).
```
AP-Hunt, line 455: This isn't quite right. Looking at the TLS ciphers on `(london.)cloudapps.digital` will show you those in use by our ALB. You need to test it against your custom domain with CloudFront to find the ones being used in communication with the client.

cadmiumcat: Is that also the case for the TLS version? 🤔

AP-Hunt: Yes

source/documentation/deploying_services/use_a_custom_domain.erb
```diff
@@ -450,6 +450,10 @@ The above example configures the CDN route to have a default time-to-live of 15
 
 The CDN route service forwards cookie headers your app by default, so cookie-based authentication will work as expected. Other headers, such as HTTP auth, are stripped by default. If you need a different configuration, see the guidance on [forwarding headers](/deploying_services/use_a_custom_domain/#forwarding-headers).
 
+### Verifying protocols and cyphers
+
+You can find the TLS version and cyphers used on GOV.UK PaaS by using an SSL testing tool such as [SSL Labs](https://www.ssllabs.com/ssltest/). For your apps, use your CloudFront custom domain as the hostname.
```
AP-Hunt, line 455: s/use your CloudFront custom domain/use your custom domain/

AP-Hunt (changes requested):
I think adding the word CloudFront here introduces some possible confusion between the tenant's domain name, and the host name of the CloudFront distribution.

source/documentation/deploying_services/use_a_custom_domain.erb
cadmiumcat: ah yes, I can see that

AP-Hunt (approved):

PR #443, opened 2022-03-29
Remove mentions of Elasticsearch

kr8n3r:
What
----

Remove mentions of Elasticsearch


cadmiumcat (commented):
Is it worth mentioning that we migrated off Elasticsearch? 
I'm thinking about people who will be thinking "I'm sure there were docs somewhere here..." and may have forgotten about it

kr8n3r:
not sure of the benefit. anyone who has/had an ES instance has been told 4 or 5 times about migration
going forward no one will have an ES instances and ES won't be present on cf marketplace.

cadmiumcat (approved):

PR #461, opened 2022-09-27
Update commonmarker  from 0.23.5 to 0.23.6

kr8n3r:
What
----

Update commonmarker from 0.23.5 to 0.23.6 to address [CVE](https://github.com/alphagov/paas-tech-docs/security/dependabot/14)


schmie (approved):
👍 local `bundle exec middleman server` is happy

PR #465, opened 2022-11-23
Update to cf v8

cadmiumcat:
What
----

We're using cf cli v8 so the examples in our docs should use the same version. 

I've gone through all the examples in the docs to check if any of the command differences from the [cf cli v7](https://docs.cloudfoundry.org/cf-cli/v7.html#differences) and [cf cli v8](https://docs.cloudfoundry.org/cf-cli/v8.html#differences) were relevant.

The removal of the `--hostname` flag also required a change in the [paas-prometheus-exporter](https://github.com/alphagov/paas-prometheus-exporter/pull/47)

How to review
-------------
1. do the changes make sense? in particular the `--hostname` flag one

1. …

Who can review
--------------

Anyone

### DIFF
--- a/source/documentation/monitoring_apps/metrics.md
+++ b/source/documentation/monitoring_apps/metrics.md
@@ -34,14 +34,10 @@ We recommend this GOV.UK PaaS account:
 1. [Push the PaaS Prometheus exporter app](/deploying_apps.html#deploying-public-apps) to Cloud Foundry without starting the app:
 
 	```
-	cf push --no-start prometheus-exporter --hostname prometheus-exporter-ORGNAME
+	cf push --no-start prometheus-exporter --var orgname=ORGNAME
 	```
 
-	where `ORGNAME` is the name of your org. For example:
-
-	```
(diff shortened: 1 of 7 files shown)


source/documentation/monitoring_apps/metrics.md
```diff
@@ -34,14 +34,10 @@ We recommend this GOV.UK PaaS account:
...
-	cf push --no-start prometheus-exporter --hostname prometheus-exporter-ORGNAME
+	cf push --no-start prometheus-exporter --var orgname=ORGNAME
 	```
 
-	where `ORGNAME` is the name of your org. For example:
-
-	```
-	cf push --no-start prometheus-exporter --hostname prometheus-exporter-exampleorg
-	```
+	where `ORGNAME` is the name of your org. Note this replaces the `--hostanme` flag from cf CLI v6, which was removed in cf CLI v7.
```
jackjoy-gds, line 40: typo on --hostanme

cadmiumcat: Good spot

cmcnallygds (approved):

PR #474, opened 2022-12-28
cdn-route services: add note on cloudfront distribution id retrieval, use `--params` argument to retrieve instance config

risicle:
What
----

https://www.pivotaltracker.com/story/show/183923174

Lifted `--params` cli invocation wording directly from redis section.

Added distribution id information as a simple note to an existing question/section rather than a whole new question/section that could get us bogged down in detail for something that's quite obscure.

### DIFF
--- a/source/documentation/deploying_services/use_a_custom_domain.erb
+++ b/source/documentation/deploying_services/use_a_custom_domain.erb
@@ -430,6 +430,7 @@ Example output:
 ```
 {
    "cache_ttl": 0,
+   "cloudfront_distribution_id": "ENHBRDSSQXJV0",
    "cloudfront_domain": "d3j6yjt78pkdqf.cloudfront.net",
    "dns_records": [
       {
@@ -446,6 +447,8 @@ Example output:
    ]
 }
 ```
+
+This also exposes information such as the CloudFront "distribution id" of the service instance, which can be useful when migrating a domain away from GOV.UK PaaS.
 
 ## How custom domains work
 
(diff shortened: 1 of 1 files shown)


corlettb (approved):

risicle:
(this needs to wait for the feature to actually go out)

source/documentation/deploying_services/use_a_custom_domain.erb
```diff
@@ -447,6 +448,8 @@ Example output:
 }
 ```
 
+This also exposes information such as the CloudFront "distribution id" of the service instance which can be useful when migrating a domain away from GOV.UK PaaS.
```
cmcnallygds, line 451:
```suggestion
This also exposes information such as the CloudFront "distribution id" of the service instance, which can be useful when migrating a domain away from GOV.UK PaaS.
```

source/documentation/deploying_services/use_a_custom_domain.erb
```diff
@@ -417,10 +417,10 @@ Server error, status code: 409, error code: 60016, message: An operation for ser
 This happens because you cannot do anything to a service instance while it's in a pending state. A CDN service instance stays pending until it detects the CNAME or ALIAS record. If this causes a problem for you, [contact support](https://www.cloud.service.gov.uk/support) to ask us to manually delete the pending instance.
```
cmcnallygds, line 417:
```suggestion
This happens because you cannot do anything to a service instance while it is in a pending state. A CDN service instance stays pending until it detects the CNAME or ALIAS record. If this causes a problem for you, [contact support](https://www.cloud.service.gov.uk/support) to ask us to manually delete the pending instance.
```

cmcnallygds (changes requested):
Suggesting a small style change

cmcnallygds (approved):

PR #481, opened 2023-01-05
Enforce a "-" separator in dependabot branch names

AP-Hunt:
What
----

Our CI system doesn't like the "/" separator in branch names. Forcing the use of a dash should fix it.

How to review
-------------

Is the dependabot config right? Is the frequency OK?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+---
+version: 2
+updates:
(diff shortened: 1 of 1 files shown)


kr8n3r (approved):

kr8n3r (commented):
make it daily

kr8n3r (approved):

PR #523, opened 2023-08-17
BAU - Fix issue with security.txt not displaying as HTML

nimalank7:
Description:
----
- Both the `security.txt` and `.well-known/security.txt` display as `.txt` files in GitHub pages not as HTML. There isn't a known way to change the MIME type to text/html in GitHub pages so the workaround involves copying the `security.txt.html` file manually
- The redirect is added for middleman server
- The empty `.nojekyll` file is added to turn off Jekyll in GitHub pages as this causes it not to recognize directories starting with . - see [here](https://github.com/orgs/community/discussions/22227)


malcgds (approved):
LGTM

PR #524, opened 2023-08-17
BAU - Use make build step in deploy.yml

nimalank7:
Description:
---
- Incorrectly uses `bundle exec middleman build` and not `make build`
- Change allows `googlec7239f490e1990a5.html` to be copied into the directory


corlettb (approved):
LGTM

PR #525, opened 2023-08-18
Update s3 bucket presigned url topic

corlettb:
What
----

Update s3 bucket presigned url topic to include a line about using them outside the PaaS

How to review
-------------

### DIFF
--- a/source/documentation/deploying_services/s3.md
+++ b/source/documentation/deploying_services/s3.md
@@ -199,6 +199,14 @@ You can [generate a presigned URL](https://docs.aws.amazon.com/AmazonS3/latest/u
 
 You can generate a presigned URL in your application using the AWS SDK package for your application's programming language. You can also generate them with the AWS CLI tool by using the [S3 bucket credentials from a service key](/deploying_services/s3/#connect-to-an-s3-bucket-from-outside-of-the-gov-uk-paas).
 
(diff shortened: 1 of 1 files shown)


source/documentation/deploying_services/s3.md
```diff
@@ -199,6 +199,12 @@ You can [generate a presigned URL](https://docs.aws.amazon.com/AmazonS3/latest/u
 
 You can generate a presigned URL in your application using the AWS SDK package for your application's programming language. You can also generate them with the AWS CLI tool by using the [S3 bucket credentials from a service key](/deploying_services/s3/#connect-to-an-s3-bucket-from-outside-of-the-gov-uk-paas).
 
+If you intend to use presigned URLs from outside of the GOV.UK PaaS (e.g. directly from the users browser) you will need to bind your service with:
```
nimalank7, line 202: Change to `

You need to allow external access to the bucket if you intend to use a presigned URL from outside of the GOV.UK PaaS.

Run the following in the command line to bind the S3 bucket to your app and allow external access:

nimalank7 (changes requested):
Change to

```
You need to allow external access to the bucket if you intend to use a presigned URL from outside of the GOV.UK PaaS.

Run the following in the command line to bind the S3 bucket to your app and allow external access:
```

source/documentation/deploying_services/s3.md
corlettb: I've updated but included an extra bit about an example.

nimalank7 (approved):
LGTM!

